PR #7, opened 2023-09-05
Fix requirements.txt

cigamit:
Fixing a change for requirement.txt that got reversed before commit


gregsowell (approved):
Approved version updates

PR #23, opened 2024-06-05
Devel

cigamit:
Fix CVE-2024-24680 - Django DOS
Upgrade packages to fix a few other CVEs. I don't believe these are exploitable, but they still show up in the scan results


gregsowell (commented):
reviewed

gregsowell (approved):
approved

gregsowell (approved):
Approve again

PR #39, opened 2025-01-29
Update main with latest devel

cigamit:
This is pulling in some other commits that were in main already, since we had to merge main back into devel.  Shouldn't affect anything as the code already exists there.


cigamit:
Actually, lets not merge this.  Need to check why on earth this has old reverted code from testing upgrading setuptools to 70.0.0

cigamit:
Ok, much better

PR #45, opened 2025-01-30
Udating for setuptools to address CVE-2024-6345

TheWitness:
##### SUMMARY
* Resolve issues with setuptools.  Addresses CVE-2024-6345
* Introduces deprecation warnings around pkg_resources which will eventually have to be replace.


zorrinna (approved):

TheWitness:
Okay, after discussing with @cigamit, we will merge this, and then conduct an investigation related to the pck_resources module.

PR #63, opened 2025-04-08
Url

zorrinna:
##### SUMMARY
Changing event handler so instead of relative path returns full URL 

##### ISSUE TYPE

##### COMPONENT NAME

 - Collection

##### ASCENDER VERSION

```
24.0.4
```

##### ADDITIONAL INFORMATION
Having relative path instead of full URL becomes an issue when running multiple ascenders and need to get more details per event 

```

```

### DIFF
--- a/awx/main/models/events.py
+++ b/awx/main/models/events.py
@@ -71,11 +71,11 @@ def emit_event_detail(event):
         InventoryUpdateEvent: 'inventory_update_id',
         SystemJobEvent: 'system_job_id',
     }[cls]
-    url = ''
+    url = event.get_absolute_url()
     if isinstance(event, JobEvent):
-        url = '/api/v2/job_events/{}'.format(event.id)
+        url += '/api/v2/job_events/{}'.format(event.id)
     if isinstance(event, AdHocCommandEvent):
-        url = '/api/v2/ad_hoc_command_events/{}'.format(event.id)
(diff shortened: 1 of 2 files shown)


cigamit:
Should rebase this against devel instead of main.

zorrinna:
Rebased against devel - Thank you

awx/main/models/events.py
```diff
@@ -71,11 +71,11 @@ def emit_event_detail(event):
         InventoryUpdateEvent: 'inventory_update_id',
         SystemJobEvent: 'system_job_id',
     }[cls]
-    url = ''
+    url = event.get_absolute_url()
     if isinstance(event, JobEvent):
-        url = '/api/v2/job_events/{}'.format(event.id)
+        url += '/api/v2/job_events/{}'.format(event.id)
```
TheWitness, line 76: Has this been tested?  Appending the URL does not seem applicable.

TheWitness (approved):
I found so many definitions of get_absolute_url() that all looked about the same.  For me to comment on this, I would have to test the function as it's not too clear to me if it's generating just 'https://some.dns.com' or 'http://some.dns.com' or something else entirely different.

awx/main/models/events.py
cigamit: I am building and about to test right now

zorrinna:
not needed

Issue #51, opened 2025-02-05
Unable to upload more than 1500 packages to ledger due to rsyslog defaults


TheWitness:
### Please confirm the following

- [x] I have checked the [current issues](https://github.com/ctrliq/ascender/issues) for duplicates.
- [x] I understand that Ascender is open source software provided for free and that I might not receive a timely response.

### Bug Summary

We get an rsyslog error 90 when a host with more than 1600 packages fails.

### Ascender version

latest

### Select the relevant components

- [ ] UI
- [ ] UI (tech preview)
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [x] Other

### Installation method

kubernetes

### Modifications

no

### Ansible version

2.14.17

### Operating system

Rocky9

### Steps to reproduce

* Attempt to gather facts for a host with 2100 packages installed
* Go to Ledger and note that the gather succeeded, it receives facts and services, but no packages

### Expected results

* All facts, services, and packages are transferred properly

### Actual results

* Packages are missed due to lack message size error

TheWitness:
The variable that needs to be increased is:

$MaxMessageSize 700000

Right now, for whatever reason, that setting is not making it into rsyslog.conf on the tasker.

TheWitness:
The real answer was to switch to the imptcp input module using a stream socket file over a datagram socket file.

PR #65, opened 2025-04-08
Fix Issue #51 - Unable to upload more than 1500 packages to ledger due to rsyslog defaults

TheWitness:
##### SUMMARY

This change updates rsyslog to use the imptcp input module over the legacy socket input module.  It does this to avoid Message too long errors (Errno 90) that occur with large packet sizes.

##### ISSUE TYPE

##### COMPONENT NAME
 - Other - Logging


zorrinna:
How did you end up testing it?

TheWitness:
The only way to test is with the new image.  So, we need to publish the images first.

cigamit (approved):

PR #67, opened 2025-04-08
Ledger authentication

cigamit:
Add an authorization header for Ledger if a token is set. This will allow us to change ledger to use tokens instead of setting trusted servers.


zorrinna (approved):

cigamit:
After setting a token/password in Ascender, you will now see a header like so
`    [HTTP_AUTHORIZATION] => Ledger thisismytoken`

So we will start looking for this in Ledger shortly.

TheWitness (approved):

PR #103, opened 2025-07-01
Add setting to disable gradient (Customer Request)

cigamit:
A customer requested the ability to disable the gradient.  This is so that transitioning internal "customers" from their old product to Ascender makes it looks close to the same.

So a setting has been added under UI Settings.  We still have a few other css settings we want to supply, so I opted for 2 separate css files.  We also move to linking to static files instead of importing, as it has to be done after the App has initialized.  This is ofcourse off by default.


awx/ui/src/components/AppContainer/AppContainer.js
```diff
@@ -53,6 +53,18 @@ function AppContainer({ navRouteConfig = [], children }) {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [config.analytics_status]);
 
+  // Dynamically load ascender.css based on ASCENDER_DISABLE_CSS UI setting from config
+  useEffect(() => {
+    const link = document.createElement('link');
```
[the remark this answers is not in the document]
cigamit: This throws a warning that link is never reassigned, so use const instead.

awx/ui/src/screens/Setting/shared/data.allSettingOptions.json
```diff
@@ -5129,6 +5138,15 @@
               "category_slug": "ui",
               "defined_in_file": false
           },
+          "ASCENDER_DISABLE_CSS": {
```
zorrinna: NIT: I would prefer to be `ASCENDER_DISABLE_GRADIENT` as otherwise it reads as disable all styling but that is definitely nitpick

cigamit: Ya, I changed the name of the option at the last minute, as I was disabling all the custom CSS we did, but then decided that I wanted only the gradient to be gone.  I can make this update.

PR #112, opened 2025-07-03
Shrink the Sidebar menu

cigamit:
This will shrink the menu sidebar to reduce the need to scroll so much.  

At my resolution, only Settings is off the screen.

### DIFF
--- a/awx/ui/src/ascender.css
+++ b/awx/ui/src/ascender.css
@@ -1,5 +1,10 @@
 #page-sidebar {
-  line-height: 12px !important;
+  line-height: 10px !important;
+}
+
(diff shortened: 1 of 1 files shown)


cigamit:
> Visually things are pretty good. In phone mode there is one distraction. I'll upload some screen shots. But are you familiar with this warning while launching with a machine credential? Assuming this translates into `Teams`. I have none...
> 
> image
> 
> EDIT: This is on 25.0.0. Maybe already fixed in devel. Will check shortly. Have to load up the database.

Typically means you have a - (dash) in a group name, which is a forbidden character.

PR #142, opened 2025-07-24
Npm updates

cigamit:
A few NPM updates were showing security issues even though dependabot wasn't alerting on them.  So I have been going through and testing the latest major stable version for each of the main dependencies and updating them.  Lots of bug fixes and performance improvements are also present in these new packages.


cigamit:
Note that rrule is kept at v2.7.1, as both 2.8.0 and 2.8.1 are reported as the next versions on npm, but on the source github, neither of those versions are released or even tagged.

TheWitness (approved):

PR #159, opened 2025-08-06
Migrate off react-script

cigamit:
Fixes vulnerability as it was tied to webpack-dev-server v4, and now we have full control over the configs / versions of dependencies


TheWitness (approved):
Hopefully this helps with lingui5 too.

PR #160, opened 2025-08-07
Lingui v5

cigamit:
Upgrade to Lingui v5.  Resolves GHSA-67mh-4wv8-2f99 and CVE-2025-54798
Update all lingui and django .po files
Fix translations for Constructed Inventories.  Add in a few other strings that should have been translated.
Translated a few strings manually, still ~90 or so that still need translating.

Fully tested every page, clicked every button, and resolved any errors.

[WIP] - Still need to resolve issues with tests (a lot have been broken since the Lingui v4 upgrade).

### DIFF
--- a/awx/ui/src/components/AppContainer/useWsPendingApprovalCount.test.js
+++ b/awx/ui/src/components/AppContainer/useWsPendingApprovalCount.test.js
@@ -4,6 +4,12 @@ import WS from 'jest-websocket-mock';
 import { mountWithContexts } from '../../../testUtils/enzymeHelpers';
 import useWsPendingApprovalCount from './useWsPendingApprovalCount';
 
+// Mock useThrottle to return the value immediately without throttling
+jest.mock('../../hooks/useThrottle', () => ({
+  __esModule: true,
+  default: jest.fn((val) => val),
+}));
+
 function TestInner() {
   return <div />;
 }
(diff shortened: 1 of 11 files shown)


awx/ui/src/api/models/Root.test.js
```diff
@@ -35,10 +35,10 @@ describe('RootAPI', () => {
 
     expect(mockHttp.post).toHaveBeenCalledTimes(2);
     expect(mockHttp.post.mock.calls[0]).toContainEqual(
-      'username=foo&password=bar&next=api%2Fv2%2Fconfig%2F'
+      'username=foo&password=bar&next=%2Fapi%2Fv2%2Fconfig%2F'
```
TheWitness, line 38: Does this look right?  I think it needs rework.

awx/ui/src/api/models/Root.test.js
```diff
@@ -35,10 +35,10 @@ describe('RootAPI', () => {
...
     );
     expect(mockHttp.post.mock.calls[1]).toContainEqual(
-      'username=foo&password=bar&next=baz'
+      'username=foo&password=bar&next=%2Fbaz'
```
TheWitness, line 41: Same comment here.  The `&2Fapi` seems wrong.

awx/ui/src/App.js
```diff
@@ -169,7 +168,11 @@ function App() {
   }
 
   if (isLoading) {
-    return <div>{i18n._(msg`Loading...`)}</div>;
+    return (
+      // Don't render I18nProvider until i18n is activated
+      // eslint-disable-next-line i18next/no-literal-string
+      <div>Loading...</div>
+    );
```
TheWitness, line 175: This explains a lot...

TheWitness (changes requested):

awx/ui/src/api/models/Root.test.js
cigamit: Nope, its correct.  The string is encoded on purpose. The only thing we are changing is adding a / at the beginning to stop it from failing.

awx/ui/src/App.js
cigamit: Ya, the I18nProvider has caused a ton of headachesm especially with the unit tests.  But if we removed it, then it means reworking everything.

awx/ui/src/api/models/Root.test.js
cigamit: Same

TheWitness (approved):

PR #312, opened 2026-04-20
Replace workbox-webpack-plugin to resolve CVE-2026-34043 GHSA-5c6j-r48x-rmvq

cigamit:
workbox-webpack-plugin is slow to update dependencies and release new versions.  

We only use the InjectManifest portion of it, so we can replace it with inject-manifest-plugin. 

This has the added benefit of removing an additional 46 co-dependencies and fixes multiple deprecation warnings.

Tacking on to this PR
- source-map-loader is deprecated so replace with webpack native functions
- identity-obj-proxy is a tiny package but hasn't been updated in 10 years, replace with local function
- Upgrade multiple dev dependencies


TheWitness (dismissed):
Reduces the threat surface.  Nice.

TheWitness (approved):

PR #332, opened 2026-05-11
Upstream 15970 - Revise start_fact_cache and finish_fact_cache to use JSON file

cigamit:
**Note: This upstream PR had some issues that were corrected in later PRs, so we will need to pull them in also before release.  Fixed in 16318 and we will also want to bring in 16408**

### UPSTREAM SUMMARY

This proposes a revision to our fact caching mechanics. Description of the change:

- for start_fact_cache: Instead of returning the tuple of 2 items (host-list, write-time), at the end of the method (after the loop completes), we write a JSON format file with the host list inside of it
- This JSON format file will be in the "artifacts" directory, but not inside of the host_facts directory
- for finish_fact_cache: Instead of accepting (host-list, write-time) as arguments, it obtains:
- host-list from a load of the JSON format file
- write-time as the modtime of that same file
- Corresponding change is that the "destination" function arguments is changed from the host_facts directory to the artifacts directory so that we can write and read this extra file.

### DIFF
--- a/awx/main/tasks/facts.py
+++ b/awx/main/tasks/facts.py
@@ -20,26 +20,27 @@ system_tracking_logger = logging.getLogger('awx.analytics.system_tracking')
 
 
 @log_excess_runtime(logger, debug_cutoff=0.01, msg='Inventory {inventory_id} host facts prepared for {written_ct} hosts, took {delta:.3f} s', add_log_data=True)
-def start_fact_cache(hosts, destination, log_data, timeout=None, inventory_id=None):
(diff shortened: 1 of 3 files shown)


TheWitness:
Looks like some good comments here @cigamit

cigamit:
Yes, a few of those are minor things we could change, but since we have to merge in 2 more PRs from upstream (which will also fix a few of these), it will be easier if we merge this as is first and then fix any issues afterwards.

TheWitness (dismissed):

TheWitness:
Got it.

TheWitness:
One conflict.  Get that and I'll approve again.

TheWitness (approved):

PR #335, opened 2026-05-11
Add per-user preferred language setting

blaipr:
##### SUMMARY
Adds a Preferred Language field to the user profile. Each user can select
their UI language from the Edit User page. The preference is stored in the
database (Profile model) and applied immediately on save, without a page
reload. Falls back to the browser's language when not set.

Changes:
- New `language` field on the `Profile` model + migration 0196
- `UserSerializer`: exposes `preferred_language` as a readable/writable API field
- UI: language dropdown in Edit User form (after User Type)
- UI: Preferred Language shown in User Detail view
- UI: locale activated immediately on save for own-profile edits
- UI: locale cached in localStorage to avoid flash on page reload
- Translations added for all 7 supported locales (es, fr, ja, ko, nl, zh, zu)

##### ISSUE TYPE
- New or Enhanced Feature

##### COMPONENT NAME
- API
- UI

##### ASCENDER VERSION
```
awx: 25.3.7.dev30+gba1f4f7.d20260511
```


cigamit:
Nice feature.  Copilot found a few things that need to be dug into to see if they are real issues. Also 49 python tests are failing, but they all have the same basic error.  Should also create some tests to test this feature.

Example
```
FAILED awx/main/tests/functional/test_rbac_oauth.py::TestOAuth2Token::test_can_create[2-can_access_list2] - django.db.utils.OperationalError: no such column: main_profile.language
```

You can run the python tests by exec-ing into the `tools_awx_1` docker container and running `make test`

blaipr:
Thanks for the review! Addressed all feedback in the latest push.

**Python test failures (49 × `no such column: main_profile.language`)**

Stale test DB — `--reuse-db` was keeping the schema from before the migration was added. Running with `--create-db` rebuilds from models and the column is there. Full suite result: **3441 passed, 6 skipped, 1 failed** — the one failure (`test_secret_key_regeneration.py::test_job_start_args`) is a pre-existing JSON decode issue unrelated to this change.

**Copilot items**

- **localStorage not cleared on blank preference** (`UserEdit.js`, `Config.js`): Fixed in both places. When `preferred_language` is empty or unsupported we now call `localStorage.removeItem('preferred_language')` and re-activate from `navigator.language` (with `'en'` fallback), instead of overwriting with `'en'`. This restores the documented fallback chain and lets users truly clear a previously set preference.

- **Broad `except Exception` in `to_representation`** (`serializers.py` line 999): Removed. `AutoOneToOneField` guarantees the profile row always exists, so there's nothing to catch.

- **`preferred_language` accepted any string** (`serializers.py` line 962): Added `validate_preferred_language()` backed by a `SUPPORTED_UI_LOCALES` frozenset (`en`, `es`, `fr`, `ja`, `ko`, `nl`, `zh`, `zu`, blank). Invalid locale codes now return a 400 with a clear message.

- **Stale `Profile` docstring** (`organization.py` line 177): Updated to mention the `language` field.

- **No tests for `preferred_language`** (`serializers.py` line 1055): Added 4 tests to `awx/main/tests/functional/api/test_serializers.py`: set valid locale persists to `profile.language`, clear to blank saves empty string, invalid locale raises `ValidationError`, `to_representation` returns the stored value. All 4 pass.

cigamit:
Everything looks good in my testing.  The one thing I would debate is whether admins should be able to modify other users preferred languages, but I see no reason to change it currently (a simple me.id === user.id would do if we want to do it later).

I did notice this was added in the Zulu language.  

```
#: screens/User/shared/UserForm.js:146
#: screens/User/UserDetail/UserDetail.js:76
msgid "Preferred Language"
msgstr "Ulimi Oluthandekayo"
```

That language should probably be hidden as it was originally only in there for testing purposes (which is why everything is blank).  Its been a while, but if I recall there was some code in there that broke in an a lingui upgrade. Before we removed it, if you used Zulu, it output random strings for everything, so it was easy to detect items that were not translated by lingui.

cigamit (approved):

PR #343, opened 2026-05-14
Fix: canceled jobs that never started should not set finished time

### LINKED ISSUE #3988: Jobs canceled before starting should not show a "finished" time
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Confusing UI if you cancel a job while it is still in "pending" state, caused by API behavior.

##### ENVIRONMENT
* AWX version: 4.0.0
* AWX install method: openshift, minishift, docker on linux, docker for mac, boot2docker
* Ansible version:  X.Y.Z
* Operating System:
* Web Browser:

##### STEPS TO REPRODUCE

Launch a job that is blocked by some other job, so it remains in "pending" (or send to an instance group with no capacity), and then cancel it.

##### EXPECTED RESULTS

Mostly null fields related to started, finished, etc.

The job never started and never finished.

##### ACTUAL RESULTS

The UI shows that the job finished, see 7844

Screen Shot 2019-06-03 at 8 50 50 AM

##### ADDITIONAL INFORMATION

Caused by the API fields.

```json
    "scm_type": "git",
    "scm_url": "https://github.com/ansible/test-playbooks.git",
    "scm_branch": "",
    "scm_clean": false,
    "scm_delete_on_update": false,
    "credential": null,
    "timeout": 0,
    "unified_job_template": 3838,
    "launch_type": "manual",
    "status": "canceled",
    "failed": true,
    "started": null,
    "finished": "2019-06-03T12:39:05.272329Z",
    "elapsed": 0.0,
    "job_args": "",
    "job_cwd": "",
    "job_env": {},
    "job_explanation": "",
    "execution_node": "",
    "result_traceback": "",
    "event_processing_finished": true,
    "project": 3838,
    "job_type": "check",
```

Expectation is that the `finished` field should remain null.

Previously @ryanpetrello made the suggestion

> Seems like if we really wanted to have accurate data, we'd want another column to record the "canceled_on" datestamp.

This issue also makes the argument for an independent field to track canceled time. (but we may prefer to track that as a separate enhancement)

flechoide:
## Summary

Fixes ansible/awx#3988 — Jobs canceled before starting should not show a
finished time.

## Problem

When a job is canceled while in `pending` state (e.g., blocked by another
job or queued to an instance group without capacity), the API returns:

```json
{
  "status": "canceled",
  "started": null,
  "finished": "2019-06-03T12:39:05.272329Z",
  "elapsed": 0.0
}
```

A job that never started should not have a `finished` timestamp.

## Root Cause

In `UnifiedJob.save()`, the completion timestamp is set unconditionally
whenever the status transitions to a terminal state (`successful`,
`failed`, `error`, `canceled`), without checking whether the job ever
started.

## Fix

Added a guard in `UnifiedJob.save()` to only set `self.finished = now()`
when `self.started is not None`. Jobs canceled before starting will now
have `finished = null`, `elapsed = 0.0`.

## Testing

- `test_cancel_pending_job_no_finished_time` — verifies the fix
- `test_cancel_running_job_sets_finished_time` — regression test
- `test_successful_job_still_sets_finished` — regression test

### DIFF
--- a/awx/main/models/unified_jobs.py
+++ b/awx/main/models/unified_jobs.py
@@ -873,10 +873,13 @@ class UnifiedJob(
         # Sanity check: Has the job just completed? If so, mark down its
         # completion time, and record its output to the database.
         if self.status in ('successful', 'failed', 'error', 'canceled') and not self.finished:
-            # Record the `finished` time.
-            self.finished = now()
-            if 'finished' not in update_fields:
-                update_fields.append('finished')
+            # Only record a finished time if the job actually started.
+            # Jobs canceled before starting (e.g. pending) should not
+            # have a finished timestamp. (Resolves #3988 from AWX main repo)
+            if self.started is not None:
+                self.finished = now()
+                if 'finished' not in update_fields:
+                    update_fields.append('finished')
 
         dq = decimal.Decimal('1.000')
         if self.elapsed is None:
--- a/awx/main/tests/functional/models/test_unified_job.py
+++ b/awx/main/tests/functional/models/test_unified_job.py
@@ -282,3 +282,51 @@ class TestTaskImpact:
         # recalculate task_impact
         jobs[0].task_impact = jobs[0]._get_task_impact()
         assert [job.task_impact for job in jobs] == [3, 2, 2]
+
+
+@pytest.mark.django_db
+class TestCanceledJobFinishedTime:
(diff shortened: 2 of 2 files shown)


cigamit (approved):
Thanks for PR.  Verified that code functions and resolves the issue.  All tests pass.

PR #357, opened 2026-05-22
Upgrade uuid to resolve CVE-2026-41907

cigamit:
It is brought in via webpack-dev-server > sockjs but we don't use the portion of the app that uses the code, so its safe to override the package until next major version of webpack-dev-server which drops sockjs completely.


TheWitness (approved):
Some copilot comments.

PR #360, opened 2026-05-29
Fix linting issues in codebase

cigamit:
flake8, black, and yamllint all return successful now All tests validate
Most changes are adding / removing newlines but some formatting changes, and some cleanup (removing unused imports, variables, etc...)


awx/main/management/commands/create_preload_data.py
```diff
@@ -66,27 +67,27 @@ def _handle(self):
                     )
                     o.galaxy_credentials.add(public_galaxy_credential)
 
-#                    i, _ = Inventory.objects.get_or_create(name='Demo Inventory', organization=o, created_by=superuser)
```
TheWitness, line 69: Is this some sort of bug or perf issue.  The last one too.  Just curious.

TheWitness: Maybe just debug logic.

TheWitness (approved):
Just a few comments out of curiosity mainly.  Good dusting off of the code otherwise.

awx/main/management/commands/create_preload_data.py
cigamit: This all was old logic that we removed when we first created Ascender (this PR just changes the spacing for the commenting out).  This file itself preloads "Demo" inventories, Users, etc.. which we did not want.  Instead we add things through our setup playbook if the user desires.

PR #380, opened 2026-06-11
Upgrade styled-components from 5 to 6

blaipr:
## SUMMARY
Upgrades styled-components 5.3.11 → 6.4.2. This is the first stage of modernizing the React stack (styled-components 5 does not even compile against React 18/19, so it has to move first). v6 has two breaking changes that affect this codebase, both handled here:

**1. The babel macro was removed.** ~50 files relied on `styled-components/macro` for the css-prop transform (`css="..."` on plain JSX elements). The macro's direct replacement, `babel-plugin-styled-components`, is now wired into the jest babel transform and the webpack babel-loader, so those usages keep working unchanged. Macro imports become plain imports; dead side-effect macro imports are removed. The `css={\`...\`}` template-literal usages (ActionsTd, ActionItem, Search, AdvancedSearch) are converted to explicit styled components.

**2. v6 no longer filters non-HTML props from styled DOM elements** — everything is forwarded to the DOM, producing React unknown-prop warnings (which this test suite treats as failures). Custom props on styled DOM elements are converted to transient (`$`-prefixed) props per the v6 migration guide: `ActionsGrid` (`$numActions`/`$gridColumns`), `ChipHolder` (`$isDisabled`), `SelectableItem`/`Indicator` (`$isSelected`), `JobEventLine` (`$isClickable`), `OutputWrapper` (`$cssMap`), `LinkG`/`StartG` (`$ignorePointerEvents`), `NodeG` (`$noPointerEvents`), `FormColumnLayout` (`$stacked`), HeaderRow's `Th` (`$alignRight`). Three props that were never consumed and only existed to be swallowed by v5's filtering are dropped (`isFirst` on JobEventLine, `isInline` on FormCheckboxLayout, `ouiaId` on a styled div — none ever reached the DOM in v5).

Rendering output is unchanged: every conversion preserves the exact CSS the component produced under v5.

**Independent of every other open PR** — verified conflict-free via pairwise `git merge-tree`.

## ISSUE TYPE
- New or Enhanced Feature

## COMPONENT NAME
- UI

## ASCENDER VERSION
```
```

## ADDITIONAL INFORMATION
All UI gates run against exactly this branch state (unmodified main + only this change):
```
npm --prefix awx/ui run test     # 544 suites passed (1 skipped), 2855 tests passed (19 skipped)
npm --prefix awx/ui run build    # production build succeeds
```
The unknown-prop warnings the suite surfaces are an exhaustive detector here: `setupTests.js` fails any test that logs a console error, so the passing suite demonstrates no styled component leaks props to the DOM anywhere the tests render.

Pre-existing oddity left untouched: `Project.helptext.js` uses `css={{...}}` objects on `<ul>` elements without ever having imported the macro — those styles never applied in v5 either and remain inert.

### DIFF
--- a/awx/ui/config/jest/babelTransform.js
+++ b/awx/ui/config/jest/babelTransform.js
@@ -24,6 +24,11 @@ module.exports = babelJest.createTransformer({
       },
     ],
   ],
+  plugins: [
+    // styled-components v6 removed the babel macro; the plugin provides
+    // the css-prop transform the codebase relies on.
+    require.resolve('babel-plugin-styled-components'),
+  ],
   babelrc: false,
   configFile: false,
 });
--- a/awx/ui/config/webpack.config.js
+++ b/awx/ui/config/webpack.config.js
@@ -418,6 +418,9 @@ module.exports = function (webpackEnv) {
                 ],
                 
                 plugins: [
+                  // styled-components v6 removed the babel macro; the plugin
+                  // provides the css-prop transform the codebase relies on.
+                  require.resolve('babel-plugin-styled-components'),
                   isEnvDevelopment &&
                     shouldUseReactRefresh &&
                     require.resolve('react-refresh/babel'),
--- a/awx/ui/package-lock.json
+++ b/awx/ui/package-lock.json
@@ -31,7 +31,7 @@
         "react-router-dom": "^5.3.3",
         "react-virtualized": "^9.22.6",
         "rrule": "2.8.1",
-        "styled-components": "5.3.11"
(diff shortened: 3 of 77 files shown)


cigamit (approved):
Tested and working just fine.

blaipr:
Both review comments addressed in the latest commit: `NodeG`'s cursor styling now reads a transient `$job` prop (and the JSX passes `$job`), and `NodeContents` reads `$isInvalidLinkTarget` — nothing custom is forwarded to the DOM anymore. All 119 visualizer tests pass; lint clean.

PR #385, opened 2026-06-11
Add React Testing Library migration infrastructure and first conversions

blaipr:
## SUMMARY
Stage 3 groundwork of the React stack modernization: start retiring **enzyme**, which is dead upstream and has no adapter beyond React 17 — it is the single hard blocker for ever moving to React 18/19 (495 test files depend on it).

This PR makes incremental conversion possible and establishes the pattern:

- **`testUtils/rtlContexts.js`** — `renderWithContexts()`, the React Testing Library counterpart of `mountWithContexts`: same context defaults and override mechanism (i18n, session, config, router with optional custom history), plus a ready `userEvent` instance and the `history` object in the return value.
- **`@testing-library/jest-dom` matchers loaded globally** in `setupTests.js` (suites no longer import them individually; existing per-file imports keep working).
- **Seven representative suites converted** as the pattern: ContentEmpty, ContentLoading, AlertModal, FormActionGroup, ChipGroup, ToolbarSyncSourceButton (interaction) and DeleteButton (modal open → confirm flow → async error). Where the enzyme versions only asserted wrapper length or component props, the RTL versions assert **rendered behavior** — e.g. ChipGroup's old test asserted a prop on an effectively empty render; it now renders actual chips and asserts the visible overflow text.

### Conversion recipe
| enzyme | RTL |
|---|---|
| `mountWithContexts(<X/>)` | `renderWithContexts(<X/>)` |
| `wrapper.find('button').simulate('click')` | `await user.click(screen.getByRole('button', { name }))` |
| `waitForElement(wrapper, 'X')` | `await screen.findBy...` |
| asserting `.props()` / `.length` | assert visible text/roles |
| `act(...)` wrapping | usually unnecessary |

Both helpers coexist; the remaining ~490 enzyme suites convert incrementally (screen directory at a time works well). When the last one converts, enzyme + the React 17 adapter come out of package.json — unblocking React 18.

**Independent of all open PRs** — test-only changes (no production code, no dependency changes). Note for the future: when the react-router compat migration (#392) merges, `rtlContexts` should gain the same controlled v6 Router layer its enzyme counterpart gets there — a 5-line follow-up.

## ISSUE TYPE
- New or Enhanced Feature

## COMPONENT NAME
- UI

## ASCENDER VERSION
```
```

## ADDITIONAL INFORMATION
```
npm --prefix awx/ui run test     # 544 suites passed (1 skipped), 2855 tests passed (19 skipped)
```
Production build unaffected (test files and test utilities only).

### DIFF
--- a/awx/ui/package.json
+++ b/awx/ui/package.json
@@ -150,7 +150,8 @@
   },
   "jest": {
     "roots": [
-      "<rootDir>/src"
+      "<rootDir>/src",
+      "<rootDir>/testUtils"
     ],
     "collectCoverageFrom": [
       "src/**/*.{js,jsx}",
@@ -164,7 +165,8 @@
     ],
     "testMatch": [
       "<rootDir>/src/**/__tests__/**/*.{js,jsx,ts,tsx}",
-      "<rootDir>/src/**/*.{spec,test}.{js,jsx,ts,tsx}"
+      "<rootDir>/src/**/*.{spec,test}.{js,jsx,ts,tsx}",
+      "<rootDir>/testUtils/**/*.{spec,test}.{js,jsx,ts,tsx}"
     ],
     "testEnvironment": "jsdom",
     "transform": {
--- a/awx/ui/src/components/AlertModal/AlertModal.test.js
+++ b/awx/ui/src/components/AlertModal/AlertModal.test.js
@@ -1,13 +1,18 @@
 import React from 'react';
-import { mountWithContexts } from '../../../testUtils/enzymeHelpers';
+import { screen } from '@testing-library/react';
+import { renderWithContexts } from '../../../testUtils/rtlContexts';
 
 import AlertModal from './AlertModal';
 
 describe('AlertModal', () => {
   test('renders the expected content', () => {
-    const wrapper = mountWithContexts(
-      <AlertModal title="Danger!">Are you sure?</AlertModal>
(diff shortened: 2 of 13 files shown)


blaipr:
All four review comments addressed: (1) `rtlContexts` now loads Lingui plural rules like `enzymeHelpers`, behind a guard that degrades cleanly when the Lingui 6 upgrade (#391) removes `loadLocaleData`/make-plural; (2) added `testUtils/rtlContexts.test.js` covering context defaults/overrides, router wiring and plural rendering — and in doing so discovered that `testUtils` was never in Jest's roots, so `enzymeHelpers.test.js` had **never actually run**; the roots/testMatch now include it, and the two rotted async-done tests were fixed (8 tests revived); (3) the ToolbarSyncSourceButton test queries by role + accessible name; (4) the DeleteButton mock attaches `.response` to a real Error (URL typo fixed too), so the network-error path is genuinely exercised. Full suite: 2,869 passed.

blaipr:

cigamit (approved):
Tested and working.  We do get this warning, but I can remove the file in another PR.

Snapshot Summary
 › 1 snapshot file obsolete from 1 test suite. To remove it, re-run jest with `-u`.
   ↳   • testUtils/__snapshots__/enzymeHelpers.test.jsx.snap

Issue #379, opened 2026-06-11
Flaky functional test failures in parallel runs caused by TransactionTestCase state leaking into the per-worker test DB


blaipr:
## Summary
Full parallel test runs (`py.test -n auto --dist=loadfile awx/main/tests/unit awx/main/tests/functional awx/conf/tests awx/sso/tests`) intermittently fail with a *varying* set of victims, all of which pass in isolation:

- `awx/main/tests/functional/utils/test_update_model.py::test_normal_get` (ERROR at teardown)
- `awx/main/tests/functional/commands/test_secret_key_regeneration.py::TestKeyRegeneration::test_job_start_args`
- `awx/main/tests/functional/api/test_survey_spec.py` password-default tests (sometimes a dozen at once)

Reproduced on unmodified `main` (cda0899) with unmodified dependency pins — three consecutive identical runs produced 2 failures, 1 failure, and a clean pass.

## Root cause analysis
The clearest evidence is the `test_update_model` teardown error:

```
django.db.utils.IntegrityError: The row in table 'main_job' with primary key '7' has an
invalid foreign key: main_job.project_id contains a value '1' that does not have a
corresponding value in main_project.unifiedjobtemplate_ptr_id.
```

The test itself only created job pk 6 — job 7 is **committed leftover state from earlier tests in the same xdist worker**. Django's `TestCase._post_teardown()` runs `check_constraints()` across the *entire* SQLite worker database, so any test that runs after the dirt was committed can become the victim.

The suite has four `TransactionTestCase` classes that really commit and clean up by flushing:

- `awx/main/tests/functional/models/test_workflow.py` (`TestWorkflowDAGFunctional`)
- `awx/main/tests/functional/task_management/test_capacity.py` (`TestInstanceGroupInstanceMapping`)
- `awx/main/tests/functional/commands/test_callback_receiver.py` (`TestCallbackBrokerWorker`)
- `awx/main/tests/functional/test_rbac_user.py` (`TestSysAuditorTransactional`)

Django's docs warn that `TransactionTestCase` mixed with `TestCase` in the same process leaks state unless the transactional tests run last. pytest/xdist does not order them that way, and `--dist=loadfile` assigns files to workers differently every run — which is exactly why the victim test changes from run to run and why simple two-file reproductions don't always trigger it (the polluting sequence needs the right 3+ file ordering on one worker).

## Suggested fix directions
[...]

PR #386, opened 2026-06-11
Remove EOL msrestazure, msrest and adal dependencies

blaipr:
## SUMMARY
Retires three end-of-life Microsoft packages in one move. `msrestazure` (archived) was imported in exactly one place: the Azure Key Vault credential plugin used `msrestazure.azure_cloud` to derive **four cloud-environment names** for the `cloud_name` dropdown — nothing else, not even the selected cloud's endpoints (auth already goes through `azure-identity`, which is MSAL-based). Inlining those names drops the whole chain from the lockfile:

- `msrestazure` 0.6.4 — EOL, archived
- `msrest` 0.7.1 — EOL, only required by msrestazure
- `adal` 1.2.7 — **abandoned by Microsoft since 2020** (superseded by MSAL), only required by msrestazure

No functional change to the plugin. `test_imported_azure_cloud_sdk_vars` asserted the old msrestazure object shape (including suffixes the plugin never read) and is rewritten to assert what the plugin actually exposes (choices list + default wiring). License files for the three packages are removed (`test_licenses` enforces the pairing).

**Independent of all open PRs.** Lockfile regenerated with `./updater.sh run`; the only delta is the three removals. Verified conflict-free against every other open PR via pairwise `git merge-tree`.

## ISSUE TYPE

## COMPONENT NAME
- API

## ASCENDER VERSION
```
```

## ADDITIONAL INFORMATION
Verified with the three packages **uninstalled** from the venv:
```
py.test awx/main/tests/functional/test_credential_plugins.py   # 13 passed
py.test awx/main/tests/functional/test_licenses.py             # 1 passed
py.test --create-db -n auto --dist=loadfile awx/main/tests/...  # 3463 passed
```
(The flaky failures seen during validation were the cross-test mock leak since root-caused and fixed in #388.)


blaipr:
Update on the flaky-test note in this PR's ADDITIONAL INFORMATION: the #379 flake family (survey_spec/update_model) referenced there is now root-caused and **fixed in #388** — a cross-test mock leak on `Job.objects.get` from six unit tests, not anything environmental. Nothing to change in this PR; with #388 merged the full parallel suite is deterministic (five consecutive clean 3476-passed validation runs).

cigamit (approved):

blaipr:
Both comments addressed. Worth noting the `cloud_name` gap **predates this PR** — the original backend never consumed it either (msrestazure only supplied the choices list); the field was always informational. Rather than just correcting the help text, the latest commit makes the field do what it says: `ClientSecretCredential` now receives the authority host mapped from `cloud_name` via `azure.identity.AzureAuthorityHosts` (Germany keeps its literal host since the constant is deprecated), so sovereign-cloud vaults authenticate correctly. The test now uses `next()`/assert instead of list indexing, plus parametrized coverage of the authority mapping (18 passed).

PR #388, opened 2026-06-11
Fix cross-test mock leak behind the flaky parallel test failures

### LINKED ISSUE #379: Flaky functional test failures in parallel runs caused by TransactionTestCase state leaking into the per-worker test DB

blaipr:
Merge PR #388 first — it makes full parallel Python test runs deterministic for reviewing everything else.

## SUMMARY
Fixes #379 — and the root cause turned out to be none of the suspects in the issue. Six unit tests assigned a Mock **directly onto the model manager**, with no patch machinery:

```python
task.model.objects.get = mock.Mock(return_value=job)   # task.model is the Job / AdHocCommand class
```

That permanently replaces `Manager.get` for the remainder of the pytest-xdist **worker process**. Whichever functional test files later shared that worker became the 'flaky' victims, and the leaked mock explains every observed failure family:

- **`api/test_survey_spec.py`**: `Job.objects.get(...)` returned a MagicMock → `json.loads(job.extra_vars)` failed with *'must be str … not MagicMock'* (up to 13 parametrized failures per run).
- **`utils/test_update_model.py`**: `update_model()` returned the unit test's leftover **in-memory Job built with `Project(pk=1)`**; the test saved it, inserting a row whose `project_id=1` references nothing — which Django's whole-DB `check_constraints()` at teardown reported as the mysterious *'main_job … invalid foreign key'* IntegrityError.
- **`commands/test_secret_key_regeneration.py`**: same frankenjob → garbage decrypted `start_args`.

The xdist scheduling lottery decided which files shared a worker with the leakers each run — hence the random victim set, why every victim passed in isolation, and why pairwise reproductions failed.

## How it was found
A temporary pytest plugin scanned **every model manager for Mock instance attributes after each test** and logged transitions per worker. Both leak sources (`unit/test_tasks.py`, `unit/tasks/test_jobs.py`) were caught within two full parallel runs, attributed to the exact tests.

## The fix
All six sites become `mocker.patch.object(task.model.objects, 'get', return_value=...)` — same behavior inside the test, automatically unwound by pytest-mock at teardown.

## ISSUE TYPE

## COMPONENT NAME
- API

## ASCENDER VERSION
```
```

## ADDITIONAL INFORMATION
Validation: **five consecutive** full parallel runs (`py.test --create-db -n auto --dist=loadfile awx/main/tests/unit awx/main/tests/functional awx/conf/tests awx/sso/tests`) all returned **3476 passed, 0 failed**, with the leak detector confirming no manager is ever left mocked. Before the fix, roughly two of every three such runs failed.

Note: the `TransactionTestCase` classes called out in the original issue text are innocent — their teardown flush verifiably leaves all tables empty. Test-only change; production code untouched. Independent of all open PRs.

### DIFF
--- a/awx/main/tests/unit/tasks/test_jobs.py
+++ b/awx/main/tests/unit/tasks/test_jobs.py
@@ -36,7 +36,7 @@ def private_data_dir():
 @mock.patch('awx.main.tasks.facts.finish_fact_cache')
 @mock.patch('awx.main.tasks.facts.settings')
 @mock.patch('awx.main.tasks.jobs.create_partition', return_value=True)
-def test_pre_post_run_hook_facts(mock_create_partition, mock_facts_settings, finish_fact_cache, private_data_dir, execution_environment):
+def test_pre_post_run_hook_facts(mock_create_partition, mock_facts_settings, finish_fact_cache, private_data_dir, execution_environment, mocker):
     # creates inventory_object with two hosts
     inventory = Inventory(pk=1)
     mock_inventory = mock.MagicMock(spec=Inventory, wraps=inventory)
@@ -65,7 +65,7 @@ def test_pre_post_run_hook_facts(mock_create_partition, mock_facts_settings, fin
     task = jobs.RunJob()
     task.instance = job
     task.update_model = mock.Mock(return_value=job)
-    task.model.objects.get = mock.Mock(return_value=job)
+    mocker.patch.object(task.model.objects, 'get', return_value=job)
 
     # run pre_run_hook
     task.facts_write_time = task.pre_run_hook(job, private_data_dir)
@@ -85,7 +85,7 @@ def test_pre_post_run_hook_facts(mock_create_partition, mock_facts_settings, fin
 @mock.patch('awx.main.tasks.facts.finish_fact_cache')
 @mock.patch('awx.main.tasks.facts.settings')
 @mock.patch('awx.main.tasks.jobs.create_partition', return_value=True)
-def test_pre_post_run_hook_facts_deleted_sliced(mock_create_partition, mock_facts_settings, finish_fact_cache, private_data_dir, execution_environment):
+def test_pre_post_run_hook_facts_deleted_sliced(mock_create_partition, mock_facts_settings, finish_fact_cache, private_data_dir, execution_environment, mocker):
     # creates inventory_object with two hosts
     inventory = Inventory(pk=1)
     mock_inventory = mock.MagicMock(spec=Inventory, wraps=inventory)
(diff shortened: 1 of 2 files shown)


cigamit (approved):
Thanks for tracking this down, was bugging me for a while but was so sporadic (only failed for me 1 in 10 times) that it was hard to pin down.  Ran through the tests multiple times and it has not occurred again.

PR #389, opened 2026-06-11
Build Activity Stream links for inventory source sync schedules

blaipr:
## SUMMARY
The Activity Stream description for a schedule belonging to an **inventory source sync** hit an explicit `throw new Error('activity.summary_fields to build this url not implemented yet')` and fell back to plain unlinked text — every other schedule parent (job template, workflow job template, project) gets a clickable link.

The blocker was that the inventory-source schedule route (`/inventories/inventory/<inventory_id>/sources/<id>/schedules/<schedule_id>/`) needs the **parent inventory id**, which the activity stream serializer didn't provide.

- **API**: `ActivityStreamSerializer._summarize_parent_ujt` now includes `inventory_id` when a schedule's parent unified job template is an `InventorySource` (the direct `inventory_source` summary gains it too, mirroring the existing `group`/`inventory_id` special case a few lines above).
- **UI**: `ActivityStreamDescription` builds the link from `summary_fields.inventory_source`, falling back to plain text when the id is unavailable rather than relying on a thrown exception for control flow.

Tests: a functional API test asserting `inventory_id` is serialized for an inventory-source schedule activity, and two UI tests (link built with the right href; graceful plain-text fallback for unknown parents).

**Independent of all open PRs** — no dependency or lockfile changes.

## ISSUE TYPE

## COMPONENT NAME
- API
- UI

## ASCENDER VERSION
```
```

## ADDITIONAL INFORMATION
Reproduce on main: create a schedule on an inventory source, then open the Activity Stream — the schedule entry renders as plain text instead of a link (the throw is caught upstream).
```
py.test --create-db -n auto --dist=loadfile awx/main/tests/...   # 3477 passed
npm --prefix awx/ui run lint                                      # clean
npm --prefix awx/ui run test                                      # 544 suites, 2857 passed
npm --prefix awx/ui run build                                     # succeeds
```

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -6059,6 +6059,9 @@ class ActivityStreamSerializer(BaseSerializer):
         summary_dict = copy.copy(SUMMARIZABLE_FK_FIELDS)
         # Special requests
         summary_dict['group'] = summary_dict['group'] + ('inventory_id',)
+        # the UI builds inventory-source schedule urls from the activity
+        # stream and needs the parent inventory id to do so
+        summary_dict['inventory_source'] = summary_dict['inventory_source'] + ('inventory_id',)
         for key in summary_dict.keys():
             if 'id' not in summary_dict[key]:
                 summary_dict[key] = summary_dict[key] + ('id',)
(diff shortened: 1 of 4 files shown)


cigamit (approved):
Tested and the name is now properly linked in Activity Stream.

PR #390, opened 2026-06-11
Surface an error when the webhook credential type lookup returns nothing

blaipr:
## SUMMARY
Fixes the second documented known-bug from the UI gaps list. `WebhookSubForm` looks up the `<service>_token` credential type to drive the **Webhook Credential** field, but when the API responds successfully with **no results**, the field silently never rendered (there's an explicit TODO at the lookup acknowledging this) — users selecting a webhook service got no credential field and no explanation.

- The form now renders an **inline warning Alert** when a webhook service is selected but no matching credential type could be resolved, instead of failing silently.
- Also fixes a secondary bug on the same screen: webhook-key failures rendered `<ContentError error={error} />` with the wrong (undefined) error object when only `webhookKeyError` was set.
- The new string is not yet in the locale catalogs (deliberately, to keep this PR free of whole-catalog churn); it falls back to its English source text at runtime and will be picked up by the next routine `extract-strings` run.

Worth noting: the pre-existing test suite ran **entirely in the silent-failure regime** — the auto-mocked `CredentialTypesAPI` returned no credential type and nothing asserted the lookup's presence. The new tests cover both sides: the lookup renders when a type resolves, and the warning appears (lookup absent) on empty results.

**No dependency, lockfile or locale-catalog changes.** Independent of every other open PR — verified conflict-free via pairwise `git merge-tree`.

## ISSUE TYPE

## COMPONENT NAME
- UI

## ASCENDER VERSION
```
```

## ADDITIONAL INFORMATION
```
npm --prefix awx/ui run test     # 544 suites, 2857 passed (incl. 2 new)
```

### DIFF
--- a/awx/ui/src/screens/Template/shared/WebhookSubForm.js
+++ b/awx/ui/src/screens/Template/shared/WebhookSubForm.js
@@ -56,8 +56,6 @@ function WebhookSubForm({ templateType }) {
         results = await CredentialTypesAPI.read({
           namespace: `${webhookServiceField.value}_token`,
         });
-        // TODO: Consider how to handle the situation where the results returns
-        // and empty array, or any of the other values is undefined or null (data, results, id)
       }
       return results?.data?.results[0]?.id;
     }, [webhookServiceField.value])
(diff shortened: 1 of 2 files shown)


blaipr:
All three comments addressed: both new tests now wait for the async credential-type lookup to settle (`waitForElement` on ContentLoading) instead of asserting after mount, and the success-path test mocks `CredentialsAPI.read`/`readOptions` with complete response shapes so `CredentialLookup` actually renders — which immediately surfaced the missing `count` field the silent-error state had been hiding, validating the concern. 6 passed.

blaipr:

cigamit (approved):

PR #392, opened 2026-06-11
Migrate react-router usage to v6 compat APIs (bridge + all non-route-tree files)

blaipr:
## SUMMARY
Migrates the UI to the react-router v6 compat APIs: installs the official incremental-migration bridge and converts **~180 of the 237 files** that used router APIs removed in react-router 6/7 (`Switch`, `useHistory`, `useRouteMatch`, `Redirect`, `withRouter`). Route-tree definitions stay on v5 for a final follow-up phase, after which `react-router-dom` flips to v6 and the compat package is removed.

### The bridge
`react-router-dom-v5-compat` (router v6 alongside v5, React 17 compatible) with `CompatRouter` inside the app's `HashRouter`; `testUtils/enzymeHelpers` renders a **controlled** nested v6 Router (location from the v5 context, navigator = the shared history) — no second subscription, no `act()` warnings, shallow rendering intact; `licenses/ui` entry for the new dependency.

### The migration (shared components, contexts, hooks and every screen directory)
`useHistory` → `useNavigate` • `history.location` → `useLocation()` • params-only `useRouteMatch` → `useParams()` • conditional `Redirect` → `Navigate` • `withRouter` → hooks (or dropped where the injected props were never read) • `Session`'s `history.listen` POP detection → `useNavigationType` with v5-faithful semantics • UIEdit's custom `hardReload` location field → navigation `state` • `history.go()` → `navigate(0)`.

Two correctness findings baked in:
1. **compat `navigate()` is not referentially stable** (identity changes with location) — redirect-on-result effects drop it from deps with explanatory eslint-disables; mechanically converting `[result, history]` deps refires redirects after unrelated navigations (caught by an existing cancel-flow test).
2. A `test.only` had silenced **14 Login tests since the fork's initial import**; removed, with two while-dark-rotted assertions fixed. The suite now runs 2,869 tests.

**Deliberately left on v5** for the final route-tree phase: `Switch`/`Route` definitions, `<Redirect from>` inside Switch, `match.path`/`match.url` route building, argument-form `useRouteMatch` prefix checks. After that phase, `react-router-dom` flips to v6 and the compat package is deleted.

**Fully independent of every other open PR — verified by pairwise `git merge-tree`.**

## ISSUE TYPE
- New or Enhanced Feature

## COMPONENT NAME
- UI

## ASCENDER VERSION
```
```

## ADDITIONAL INFORMATION
```
npm --prefix awx/ui run test     # 544 suites passed (1 skipped), 2869 tests passed (5 skipped)
py.test awx/main/tests/functional/test_licenses.py   # 1 passed
```
The v6-context requirement is self-enforcing in tests: a migrated component rendered without the bridge fails loudly (`useNavigate() may be used only in the context of a <Router>`).


blaipr:
Both comments addressed: `JobTemplateAdd` now checks `includes('resource_id') && includes('resource_name')` (the old expression only ever tested for `resource_name` — good catch, this also matched the pre-migration behavior so it's a genuine bug fix), and `PaginatedTable` uses a single `useLocation()` call. Affected suites pass (29 tests).

blaipr:
Both fixed. The resource param check now requires both values before building the defaults, and the duplicate useLocation call is gone.

blaipr:
Branch refreshed against current main (it was 14 merges behind and predated the RTL infrastructure from #385). Two substantive additions beyond the merge:

- **Wired the compat bridge into the RTL test harness.** `renderWithContexts` (testUtils/rtlContexts.js, added by #385) mounted only the v5 Router, so any RTL suite rendering a component this PR migrates would have thrown `useNavigate() may be used only in the context of a <Router>` after merge — zero textual conflicts, but a guaranteed runtime break against the User suites converted in #398. It now nests the same controlled v6 Router as enzymeHelpers (location from the v5 context, navigator = the shared history), with a regression test exercising the compat hooks through the harness.
- Removed the orphaned `enzymeHelpers.test.jsx.snap` that fails the run under jest's ci mode.

Full suite against current main: **546 suites passed (1 skipped), 2,893 tests passed (5 skipped)**, lint clean. Re-verified conflict-free against main and every other open PR (#396, #397, #398) — the shared rtlContexts.js edits here and in #398 use identical hunks, so they merge cleanly in either order.

cigamit (approved):

PR #391, opened 2026-06-11
Modernize the frontend toolchain: Lingui 6, ESLint 9 and @dagrejs/dagre

blaipr:
## SUMMARY
Modernizes three pieces of the frontend toolchain — **Lingui 5 → 6**, **ESLint 8 → 9 (flat config)** and the unmaintained **dagre replaced by @dagrejs/dagre**. They ship as one change because each regenerates `package-lock.json` and their lockfile entries interleave; together the lockfile is regenerated once, cleanly.

### Lingui 5 → 6
`.linguirc` → `lingui.config.js` (+`@lingui/format-po`); removed `i18n.loadLocaleData()` everywhere (API deleted in v6; plurals come from `Intl.PluralRules`) and the now-unused `make-plural` dependency; webpack rule relaxing fully-specified ESM resolution (v6 is ESM-only, React 17 has no exports map); catalogs re-extracted with translations preserved (es: 1,896 → 1,898 filled msgstr).

### ESLint 8 → 9.39 flat config
`eslint.config.mjs` replaces `.eslintrc.json`/`.eslintignore`; dead `eslint-config-airbnb`/`eslint-config-react-app` (capped at eslint ^8) replaced by `eslint-config-airbnb-extended`; import rules → import-x; i18next plugin 6 (option schema translated, shallow-merge defaults spread in); react-hooks 7 (new React-Compiler rules off — ~105 sites predate them, candidate follow-up); rules newer than the airbnb@19 baseline off for parity; `npm audit fix` clears GHSA-w7jw-789q-3m8p. **ESLint 10 remains blocked** by plugin peer ranges (react/jsx-a11y/import-x/babel-parser all ^9) and its removal of core formatting rules — the @stylistic remaps here clear that hurdle in advance.

### dagre → @dagrejs/dagre 3
The maintained continuation; one import line in `WorkflowUtils.js` (sole consumer); layout test now asserts geometry without library internals; `licenses/ui` file renamed.

**Known benign artifact:** one unused-eslint-disable warning in `StatusLabel.js` — the directive is required under #380's eslint-8 baseline, so it stays to keep #380 conflict-free; remove after both merge.

**Fully independent of every other open PR — verified by pairwise `git merge-tree` against all of them.**

## ISSUE TYPE
- New or Enhanced Feature

## COMPONENT NAME
- UI

## ASCENDER VERSION
```
awx: 25.4.1.dev5+gcda0899.d20260610
```

## ADDITIONAL INFORMATION
```
[...]

PR #393, opened 2026-06-11
Show elapsed time on the workflow job output toolbar

blaipr:
## SUMMARY
Regular job output pages show a live `hh:mm:ss` elapsed badge — ticking every second while the job runs, switching to the API's final `elapsed` value on completion. **Workflow job output pages showed no elapsed time at all**, live or final.

The workflow output toolbar now renders the same badge next to **Total Nodes**:
- a one-second interval recomputes elapsed from `job.started` while the workflow runs
- the existing websocket-driven refetch (`useWsJob` refetches on terminal statuses) delivers `finished`/`elapsed`, so the badge settles on the authoritative final value without a reload
- `calculateElapsed` moves from a private helper inside `OutputToolbar` to `util/dates` alongside `secondsToHHMMSS`, and `OutputToolbar` now uses both shared helpers (no behavior change there)

No locale-catalog changes needed — both strings (`Elapsed`, `Elapsed time that the job ran`) already exist in the catalogs.

**Independent of every other open PR — verified conflict-free via pairwise `git merge-tree`** (the elapsed-state hooks are deliberately placed clear of the hunks #392 touches in the same file).

## ISSUE TYPE
- New or Enhanced Feature

## COMPONENT NAME
- UI

## ASCENDER VERSION
```
awx: 25.4.1.dev6+g9b44420
```

## ADDITIONAL INFORMATION
Tests: unit coverage for `calculateElapsed` (zero/unstarted cases, fake-timer arithmetic) and two toolbar tests (badge ticks under fake timers while running; shows the formatted final `elapsed` and stops ticking once finished).
```
npm --prefix awx/ui run test     # 544 suites, 2859 passed (incl. 6 new)
```

### DIFF
--- a/awx/ui/package-lock.json
+++ b/awx/ui/package-lock.json
@@ -23760,7 +23760,7 @@
       "version": "4.9.5",
       "resolved": "https://registry.npmjs.org/typescript/-/typescript-4.9.5.tgz",
       "integrity": "sha512-1FXk9E2Hm+QzZQ7z+McJiHL4NW1F2EzMu9Nq9i3zAaGqibafqYwCVU6WyWAuyQRRzOlxou8xZSyXLEN8oKj24g==",
-      "devOptional": true,
+      "dev": true,
       "license": "Apache-2.0",
       "peer": true,
       "bin": {
--- a/awx/ui/package-lock.json
+++ b/awx/ui/package-lock.json
@@ -23760,7 +23760,7 @@
       "version": "4.9.5",
       "resolved": "https://registry.npmjs.org/typescript/-/typescript-4.9.5.tgz",
       "integrity": "sha512-1FXk9E2Hm+QzZQ7z+McJiHL4NW1F2EzMu9Nq9i3zAaGqibafqYwCVU6WyWAuyQRRzOlxou8xZSyXLEN8oKj24g==",
-      "dev": true,
+      "devOptional": true,
       "license": "Apache-2.0",
       "peer": true,
       "bin": {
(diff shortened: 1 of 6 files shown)


cigamit:
We should probably add a right margin or some padding to the badge.  The badge should probably also have a fixed width (or the fixed width font), as it constantly changes size as the characters get wider or slimmer.

image

cigamit:
Otherwise tested and works fine.

blaipr:
All four comments addressed: (1) the badge keeps showing the last live timer value while `finished` is set but the refetched `elapsed` hasn't arrived (`job.elapsed != null` guard, applied to both toolbars since they share the ws semantics; new test covers the window); (2)+(3) tests select badges by id instead of count/order; (4) the lockfile metadata churn is reverted — it was accidental, from a cross-branch `npm install`. 121 tests pass across the touched suites.

blaipr:

blaipr:
@cigamit good catch — and apologies, my earlier "all feedback addressed" comment was posted after yours but only covered the Copilot review items; I'd missed the conversation thread. Both points from your screenshot are fixed now:

- **spacing**: the elapsed badge has a 20px right margin, so it no longer sits flush against "Total Nodes"
- **width**: the badge has `min-width: 70px` plus `font-variant-numeric: tabular-nums`, so it keeps a constant width while the timer ticks instead of jittering with digit widths

The job output toolbar's elapsed badge got the same width treatment, since it had the identical wobble (its spacing was already fine via BadgeGroup). 132 tests pass across both toolbars; lint clean.

cigamit (approved):

PR #396, opened 2026-06-12
Wire up date filtering in the Search toolbar

blaipr:
## SUMMARY
`qsConfig` has accepted `dateFields` since the search system was built — every list's config defaults to `['modified', 'created']` — but the Search component never rendered date inputs (an explicit TODO at the fallback text input), so no list could offer date filtering.

**Search** now renders a date-specific control when the selected search column is one of the qsConfig `dateFields`:
- a native `type=date` input plus an operator select — **On or after** → `__gte`, **Before** → `__lt` (deliberately the two operators whose semantics are exact when a date-only value meets the API's datetime fields; `lte`/`gt` on a bare date are off-by-a-day traps)
- submit calls `onSearch` with the suffixed param (e.g. `created__gte=2026-06-01`); the existing leftover-chip handling displays and removes it, and `getChipsByKey` now labels such chips with the base column's name (`Created (created__gte)`) instead of the raw key

**First consumers:** the Jobs list gains **Created** and **Finished** search columns (`finished` added to its dateFields), and the Activity Stream gains a **Time** column on its `timestamp` field. Any other list opts in by adding a search column whose key is in its `dateFields`.

Closes the 'date filtering not wired up' entry from the UI known-gaps list.

## ISSUE TYPE
- New or Enhanced Feature

## COMPONENT NAME
- UI

## ASCENDER VERSION
```
awx: 25.4.1.dev11+g51188f0
```

## ADDITIONAL INFORMATION
Tests: six new Search tests (date control renders for date columns only; default-operator submit; operator switch changes the param; stale-value leak regression; Enter-key submit; plain input preserved otherwise) and a getChipsByKey labeling test.
```
npm --prefix awx/ui run test     # 545 suites, 2864 passed (incl. 7 new)
```
New UI strings follow the catalog convention (English fallback until the next routine extract). **Independent of all five open PRs — verified pairwise with `git merge-tree`.**

### DIFF
--- a/awx/ui/src/components/JobList/JobList.js
+++ b/awx/ui/src/components/JobList/JobList.js
@@ -41,7 +41,8 @@ function JobList({
       not__launch_type: 'sync',
       ...defaultParams,
     },
-    ['id', 'page', 'page_size']
+    ['id', 'page', 'page_size'],
+    ['created', 'modified', 'finished']
   );
 
   const { me } = useConfig();
@@ -247,6 +248,14 @@ function JobList({
             {
               name: t`Limit`,
               key: 'job__limit',
+            },
+            {
+              name: t`Created`,
+              key: 'created',
+            },
+            {
+              name: t`Finished`,
+              key: 'finished',
             },
           ]}
           headerRow={
--- a/awx/ui/src/components/Search/Search.js
+++ b/awx/ui/src/components/Search/Search.js
@@ -99,6 +99,11 @@ function Search({
     onShowAdvancedSearch(actualSearchKey === 'advanced');
     setIsFilterDropdownOpen(false);
     setSearchKey(actualSearchKey);
+    // a value typed for the previous key must not leak into the next one -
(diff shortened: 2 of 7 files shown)


blaipr:
Good catch, fixed: the operator dropdown state now resets when switching search columns, with a test for the switch-away-and-back case.

blaipr:
Heads-up on i18n: this PR introduces four new msgids (`On or after`, `Before`, `Date operator select`, `Date search input`) that are deliberately **not** extracted into the `.po` catalogs yet. Extracting them with main's current lingui config would re-fold the whole catalog in the old wrapped style and conflict with the catalog rework on the `language-updates` branch (new `lingui.config.ts`, `foldLength: 0` long-line format, de/it locales — see https://github.com/ctrliq/ascender/pull/391#issuecomment-4693051588). Once `language-updates` merges, an `extract-strings` run will pick these up in the new long-line format.

cigamit (approved):
Tested and working perfectly

PR #398, opened 2026-06-13
Convert the User screen test suites from Enzyme to React Testing Library

blaipr:
## SUMMARY

Converts all **22 test suites under `screens/User/`** (~1,100 lines) from Enzyme to React Testing Library, continuing the migration started in #385. Zero `enzyme` imports remain in the directory. Tests now drive the real UI through roles and labels instead of reaching into component props, so they survive the React 18 upgrade that I'm pursuing.

### Infrastructure fixes the conversion surfaced

- **`setupTests.js`**: `@nteract/mockument` replaces `document.createRange` with a stub missing `cloneRange` and most of the Range API, which crashes `@testing-library/user-event`'s pointer handling. Restored jsdom's real Range and grafted on the two rect methods the editor components actually need from the stub.
- **`testUtils/rtlContexts.js` gains `settleTooltips()`**: closing a PF Modal restores focus (on a timeout) to the button that opened it; if that button is wrapped in a PF `Tooltip`, the tooltip engages after its 300ms entry delay and a tooltip still pending at unmount logs *"state update on unmounted component"* into the **next** test — an order-dependent failure that bisects to nothing. Applied in the three list suites that end shortly after closing an error modal (`UserTeamList`, `UserList`, `UserTokenList`). Every list-screen conversion from here on will want this helper.
- **`testUtils/rtlContexts.js` gains `assertDetail()`**: shared label/value assertion for `<Detail>` pairs; `UserDetail` and `UserTokenDetail` use it now, the other ~10 detail suites will as they convert.
- **`DisassociateButton`**: PF4 `Tooltip` doesn't accept `ouiaId` and forwards it to a DOM div, which React warns about. Enzyme's `simulate('click')` never opened the tooltip so this went unnoticed for years; user-event's real hover found it. One-line removal — nothing selects on that id.
- Deleted the orphaned `enzymeHelpers.test.jsx.snap` that made Jest print "1 snapshot file obsolete" on every run.

### Gotchas for the next conversion batches

1. `resetMocks: true` strips `jest.fn(() => …)` implementations from `jest.mock` factories between tests — use a plain function (capture props via a `mock`-prefixed variable if you need them).
2. PF required-field labels render as `"Label *"` — anchor with `getByLabelText(/^Label/)`, exact match fails.
3. An `aria-label` overrides visible text as the accessible name (the Edit link is `name: 'edit'`).
4. PF Modal's self-referencing `aria-labelledby` inflates the dialog's accessible name — match dialogs with a regex, not the exact title.
5. Tests that end right after closing a modal need `settleTooltips()` (see above).
6. Run with `TZ=UTC` like the npm script does — bare `npx jest` fails the date assertions with a confusing 1-hour offset.

**Fully independent of every other open PR.**

## ISSUE TYPE

## COMPONENT NAME

- UI

## ASCENDER VERSION

```
awx: 25.4.1.dev17+g3946cc5.d20260613
```

## ADDITIONAL INFORMATION

```
npm --prefix awx/ui run test     # 546 suites passed (1 skipped), 2878 tests passed (19 skipped)
```
The converted suites plus testUtils were additionally run 3× in a loop to shake out order/timing flakes (106 tests green each pass), since the tooltip bug class this PR fixes only shows up across test boundaries.


blaipr:
Rebased onto current main now that the react-router v6 compat migration has merged. Two things the merge required:

- **`testUtils/rtlContexts.js` conflict** — both branches edited this file (the compat work wired a controlled `CompatV6Layer` into `renderWithContexts`; this branch added the `settleTooltips`/`assertDetail` helpers and dropped the dead Lingui v5 plural block). Resolved by keeping both, and removing the now-unnecessary `eslint-disable import-x/prefer-default-export` since the file exports three symbols.
- **`UserForm` controlled-input fix** — with the compat bridge live, a successful submit's `navigate()` now drives a re-render of the still-mounted form, which surfaced a latent bug: the submit handler mutated Formik's `values` object (`delete values.password`), flipping the password field from controlled to uncontrolled and tripping the `setupTests` console-error trap. Fixed by building the submit payload from a copy and leaving Formik's state intact.

Full suite green against current main: **546 suites passed (1 skipped), 2,902 tests passed (5 skipped)**, lint clean.
GitHub now reports the branch mergeable.

cigamit (approved):

PR #407, opened 2026-06-14
Convert CredentialType route tree to react-router v6 Routes

blaipr:
##### SUMMARY

Continues the react-router v5 → v6 route-tree migration. Converts the **CredentialType** screen's route trees from the v5 `<Switch>`/`<Route>`/`<Redirect>` API to v6 `<Routes>`/`<Route>` via `react-router-dom-v5-compat`, following the exact pattern established for the Application route tree.

- `CredentialTypes.js` (list router): `<Switch>` → `<Routes>`; child routes use the `element` prop; the detail route is `/credential_types/:id/*` so the nested `<CredentialType>` tree can match the rest of the path.
- `CredentialType.js` (detail router): `<Switch>` → `<Routes>` with **relative** child paths (`edit`, `details`); the `exact` `<Redirect>` is replaced by an index route that `<Navigate replace>`s to `details`.
- Both test suites rewritten with `renderWithContexts` (RTL) to assert **real v6 route resolution** (which panel renders per URL, index redirect, 404 error) instead of enzyme structural checks.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

- Part of the incremental react-router v6 migration; each screen's route tree is independent and converted on its own branch.
- Tests: 8 route-resolution tests across the two converted suites; full `screens/CredentialType` directory passes (8 suites / 37 tests).

### DIFF
--- a/awx/ui/src/screens/CredentialType/CredentialType.js
+++ b/awx/ui/src/screens/CredentialType/CredentialType.js
@@ -1,12 +1,12 @@
 import React, { useEffect, useCallback } from 'react';
+import { Link } from 'react-router-dom';
 import {
-  Link,
-  Redirect,
+  Routes,
   Route,
-  Switch,
+  Navigate,
   useLocation,
   useParams,
-} from 'react-router-dom';
+} from 'react-router-dom-v5-compat';
 
 import { useLingui } from '@lingui/react/macro';
 import { Card, PageSection } from '@patternfly/react-core';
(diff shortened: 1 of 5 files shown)


blaipr:
Thanks for the review. Addressed in `a3e40cfe56`: fixed the doubled `/* /*` in the nested-route comment, removed the unused `CredentialTypes` API mock from the dispatcher test, and moved `Link` onto `react-router-dom-v5-compat` so the component uses a single router source.

cigamit (approved):

PR #408, opened 2026-06-14
Convert NotificationTemplate route tree to react-router v6 Routes

blaipr:
##### SUMMARY

Continues the react-router v5 → v6 route-tree migration. Converts the **NotificationTemplate** screen's route trees from the v5 `<Switch>`/`<Route>`/`<Redirect>` API to v6 `<Routes>`/`<Route>` via `react-router-dom-v5-compat`, following the same pattern used for the Application and CredentialType route trees.

- `NotificationTemplates.js` (list router): `<Switch>` → `<Routes>`; child routes use the `element` prop; the detail route is `/notification_templates/:id/*` so the nested `<NotificationTemplate>` tree can match the rest of the path; the unused `useRouteMatch` `match.url` is replaced by explicit paths.
- `NotificationTemplate.js` (detail router): `<Switch>` → `<Routes>` with **relative** child paths (`edit`, `details`); the `exact` `<Redirect>` becomes an index route that `<Navigate replace>`s to `details`; the Details tab link is built from the route id instead of `match.url`.
- Tests: the list test is converted to RTL and a new detail test is added, both using `renderWithContexts` to assert **real v6 route resolution** (which panel renders per URL, index redirect, 404 error).

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

- Tests: 7 route-resolution tests across the two suites; full `screens/NotificationTemplate` directory passes (6 suites / 31 tests).

### DIFF
--- a/awx/ui/src/screens/NotificationTemplate/NotificationTemplate.js
+++ b/awx/ui/src/screens/NotificationTemplate/NotificationTemplate.js
@@ -2,15 +2,14 @@ import React, { useEffect, useCallback } from 'react';
 import { useLingui } from '@lingui/react/macro';
 import { Card, PageSection } from '@patternfly/react-core';
 import { CaretLeftIcon } from '@patternfly/react-icons';
+import { Link } from 'react-router-dom';
 import {
-  Link,
-  Switch,
+  Routes,
   Route,
-  Redirect,
+  Navigate,
   useParams,
-  useRouteMatch,
   useLocation,
-} from 'react-router-dom';
+} from 'react-router-dom-v5-compat';
 import useRequest from 'hooks/useRequest';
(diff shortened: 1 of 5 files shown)


blaipr:
Thanks for the review. Addressed in `f417913506` and `c20c3e6818`: fixed the doubled `/* /*` comment and dropped the unused API mock; and reworked the load so the bare `/:id` no longer fetches before redirecting. The index `<Navigate>` now renders unconditionally and the detail/edit routes show a loading fallback, so the bare URL is skipped and only `/details` fetches. Real navigation (edit to details) still re-fetches, so saved changes are reflected.

cigamit (approved):

PR #409, opened 2026-06-14
Convert Team route tree to react-router v6 Routes

blaipr:
##### SUMMARY

Continues the react-router v5 → v6 route-tree migration. Converts the **Team** screen's route trees from the v5 `<Switch>`/`<Route>`/`<Redirect>` API to v6 `<Routes>`/`<Route>` via `react-router-dom-v5-compat`, following the pattern used for the Application, CredentialType and NotificationTemplate route trees.

- `Teams.js` (list router): `<Switch>` → `<Routes>`; child routes use the `element` prop; the detail route is `/teams/:id/*` so the nested `<Team>` tree can match the rest of the path.
- `Team.js` (detail router): `<Switch>` → `<Routes>` with **relative** child paths (`details`, `edit`, `access`, `roles`); the `exact` `<Redirect>` becomes an index route that `<Navigate replace>`s to `details`; the catch-all not-found route stays as `path="*"`.
- Tests: both suites rewritten with `renderWithContexts` (RTL) to assert **real v6 route resolution** — which panel renders per URL, the index redirect, the unknown-sub-route not-found error, and the 404 detail error.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

- Tests: 10 route-resolution tests across the two converted suites; full `screens/Team` directory passes (10 suites / 45 tests).

### DIFF
--- a/awx/ui/src/components/UserAndTeamAccessAdd/UserAndTeamAccessAdd.js
+++ b/awx/ui/src/components/UserAndTeamAccessAdd/UserAndTeamAccessAdd.js
@@ -30,11 +30,15 @@ function UserAndTeamAccessAdd({
   apiModel,
   onClose,
   onError,
+  resourceId,
 }) {
   const { t } = useLingui();
   const [selectedResourceType, setSelectedResourceType] = useState(null);
   const [stepIdReached, setStepIdReached] = useState(1);
-  const { id: userId } = useParams();
+  const { id: routeId } = useParams();
+  // The caller passes the resource id explicitly (works whether the parent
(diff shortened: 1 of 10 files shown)


blaipr:
Thanks for the review. Fixed the doubled `/* /*` in the nested-route comment in `6d47dd9a46`.

cigamit:
When adding a role to a Team
Error: Failed to associate role
django.request Not Found: /api/v2/teams/undefined/roles/

cigamit (approved):
Role association appears to be working now

PR #410, opened 2026-06-14
Convert ExecutionEnvironment route tree to react-router v6 Routes

blaipr:
##### SUMMARY

Continues the react-router v5 → v6 route-tree migration. Converts the **ExecutionEnvironment** screen's route trees from the v5 `<Switch>`/`<Route>`/`<Redirect>` API to v6 `<Routes>`/`<Route>` via `react-router-dom-v5-compat`, following the pattern used for the Application, CredentialType, NotificationTemplate and Team route trees.

- `ExecutionEnvironments.js` (list router): `<Switch>` → `<Routes>`; child routes use the `element` prop; the detail route is `/execution_environments/:id/*` so the nested `<ExecutionEnvironment>` tree can match the rest of the path.
- `ExecutionEnvironment.js` (detail router): `<Switch>` → `<Routes>` with **relative** child paths (`edit`, `details`, `templates`); the `exact` `<Redirect>` becomes an index route that `<Navigate replace>`s to `details`.
- Tests: the list test is converted to RTL and a new detail test is added, both using `renderWithContexts` to assert **real v6 route resolution** (which panel renders per URL, index redirect, 404 error).

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

- Tests: 8 route-resolution tests across the two suites; full `screens/ExecutionEnvironment` directory passes (10 suites / 51 tests).

### DIFF
--- a/awx/ui/src/screens/ExecutionEnvironment/ExecutionEnvironment.js
+++ b/awx/ui/src/screens/ExecutionEnvironment/ExecutionEnvironment.js
@@ -1,12 +1,12 @@
 import React, { useEffect, useCallback } from 'react';
+import { Link } from 'react-router-dom';
 import {
-  Link,
-  Redirect,
+  Routes,
   Route,
-  Switch,
+  Navigate,
   useLocation,
   useParams,
-} from 'react-router-dom';
+} from 'react-router-dom-v5-compat';
 
 import { useLingui } from '@lingui/react/macro';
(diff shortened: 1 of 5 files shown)


blaipr:
Thanks for the review. Fixed the doubled `/* /*` in the nested-route comment in `768410eae6`.

cigamit (approved):

PR #411, opened 2026-06-14
Convert Organization route tree to react-router v6 Routes

blaipr:
##### SUMMARY

Continues the react-router v5 → v6 route-tree migration. Converts the **Organization** screen's route trees from the v5 `<Switch>`/`<Route>`/`<Redirect>` API to v6 `<Routes>`/`<Route>` via `react-router-dom-v5-compat`, following the pattern used for the Application, CredentialType, NotificationTemplate, Team and ExecutionEnvironment route trees.

- `Organizations.js` (list router): `<Switch>` → `<Routes>`; child routes use the `element` prop; the detail route is `/organizations/:id/*` so the nested `<Organization>` tree can match the rest of the path; the unused `useRouteMatch` `match.path` is replaced by explicit paths.
- `Organization.js` (detail router): `<Switch>` → `<Routes>` with **relative** child paths (`details`, `edit`, `access`, `teams`, `notifications`, `execution_environments`); the `exact` `<Redirect>` becomes an index route that `<Navigate replace>`s to `details`; the catch-all not-found route stays as `path="*"`; `match.url` / `match.params.id` are replaced with the `organizationId` route param; a stray `,` text node after the not-found route is removed.
- Tests: both suites rewritten with `renderWithContexts` (RTL) to assert **real v6 route resolution** — each tab panel, the notifications tab for an auditor, the index redirect, the unknown-sub-route not-found error, and the 404 detail error.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

- Tests: 12 route-resolution tests across the two converted suites; full `screens/Organization` directory passes (12 suites / 66 tests).

### DIFF
--- a/awx/ui/src/screens/Organization/Organization.js
+++ b/awx/ui/src/screens/Organization/Organization.js
@@ -1,14 +1,13 @@
 import React, { useCallback, useEffect, useRef } from 'react';
 import { useLingui } from '@lingui/react/macro';
+import { Link } from 'react-router-dom';
 import {
-  Switch,
+  Routes,
   Route,
-  Redirect,
-  Link,
+  Navigate,
   useLocation,
   useParams,
-  useRouteMatch,
-} from 'react-router-dom';
+} from 'react-router-dom-v5-compat';
 import { CaretLeftIcon } from '@patternfly/react-icons';
 import { Card, PageSection } from '@patternfly/react-core';
 import useRequest from 'hooks/useRequest';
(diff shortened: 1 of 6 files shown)


blaipr:
Thanks for the review. Fixed the doubled `/* /*` in the nested-route comment in `af9a898187`.

cigamit (approved):

PR #412, opened 2026-06-14
Convert InstanceGroup route tree to react-router v6 Routes

blaipr:
##### SUMMARY

Continues the react-router v5 → v6 route-tree migration. Converts the **InstanceGroup** screen's route trees — instance groups, container groups, and the nested per-group instances subtree — from the v5 `<Switch>`/`<Route>`/`<Redirect>` API to v6 `<Routes>`/`<Route>` via `react-router-dom-v5-compat`, following the pattern used for the earlier screen conversions.

- `InstanceGroups.js` (list router): `<Switch>` → `<Routes>`; child routes use the `element` prop; the container-group and instance-group detail routes use a trailing `/*` so their nested trees match the rest of the path. v6 route ranking keeps the static `container_group` / `add` segments ahead of the `:id` params.
- `InstanceGroup.js` / `ContainerGroup.js` (detail routers): `<Switch>` → `<Routes>` with **relative** child paths; the `exact` `<Redirect>` becomes an index route that `<Navigate replace>`s to `details`. InstanceGroup delegates its instances tab with `instances/*`.
- `Instances/Instances.js` (nested subtree): `<Switch>` → `<Routes>` with relative paths — the index renders the instance list, `:instanceId/details` renders the instance detail, and `:instanceId` redirects to details.
- Tests: the three detail/list suites are rewritten with `renderWithContexts` (RTL) and a new `Instances` route test is added, all asserting **real v6 route resolution** (each panel, the index redirects, and the 404 errors).

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

- This is the first conversion with a nested in-screen route subtree (`Instances`); it converts parent and child together so the nested routes resolve correctly under the v6 parent.
- Tests: 19 route-resolution tests across the four converted suites; full `screens/InstanceGroup` directory passes (17 suites / 86 tests).

### DIFF
--- a/awx/ui/src/screens/InstanceGroup/ContainerGroup.js
+++ b/awx/ui/src/screens/InstanceGroup/ContainerGroup.js
@@ -1,12 +1,12 @@
 import React, { useEffect, useCallback } from 'react';
+import { Link } from 'react-router-dom';
 import {
-  Link,
-  Redirect,
+  Routes,
   Route,
-  Switch,
+  Navigate,
   useLocation,
   useParams,
-} from 'react-router-dom';
+} from 'react-router-dom-v5-compat';
 
 import { useLingui } from '@lingui/react/macro';
 import { CaretLeftIcon } from '@patternfly/react-icons';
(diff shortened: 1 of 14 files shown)


blaipr:
Thanks for the review. Addressed in `fac476528e`: added a `path="*"` not-found fallback to InstanceGroup and ContainerGroup so unknown sub-paths render a ContentError instead of a blank panel, and fixed the doubled `/* /*` comment.

cigamit (approved):

PR #413, opened 2026-06-14
Convert Instances route tree to react-router v6 Routes

blaipr:
##### SUMMARY

Continues the react-router v5 → v6 route-tree migration. Converts the **Instances** screen's route trees from the v5 `<Switch>`/`<Route>`/`<Redirect>` API to v6 `<Routes>`/`<Route>` via `react-router-dom-v5-compat`, following the pattern used for the earlier screen conversions.

- `Instances.js` (list router): `<Switch>` → `<Routes>`; child routes use the `element` prop; the detail route is `/instances/:id/*` so the nested `<Instance>` tree can match the rest of the path. The `/instances/:id/edit` route stays a sibling of the detail route — v6 route ranking keeps the static `edit` segment ahead of the `:id/*` splat.
- `Instance.js` (detail router): `<Switch>` → `<Routes>` with **relative** child paths (`details`, and the K8s-only `listener_addresses` / `peers`); the `exact` `<Redirect>` becomes an index route that `<Navigate replace>`s to `details`; the catch-all not-found route stays as `path="*"`; `match.url` / `match.params.id` are replaced with the `id` route param.
- Tests: new `Instance` and `Instances` route-resolution suites with `renderWithContexts` (RTL) — each tab panel, the K8s-gated peers/listener tabs, the index redirect, the edit sibling route, and the not-found error.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

- These two route trees had no top-level test files before; this adds 9 route-resolution tests. Full `screens/Instances` directory passes (9 suites / 44 tests).


blaipr:
Thanks for the review. Addressed in `ea6ecf4b46`: fixed the doubled `/* /*` comment and renamed the peers test to match what it asserts (listener addresses are covered by their own test).

cigamit (approved):
Tested what I could with this in my dev environment. Since I am unable to create instances there, will approve and let the image build, and will test the resulting image in my demo environment to verify all the instance routes are working.

PR #414, opened 2026-06-14
Convert Credential route tree to react-router v6 Routes

blaipr:
##### SUMMARY

Continues the react-router v5 → v6 route-tree migration. Converts the **Credential** screen's route trees from the v5 `<Switch>`/`<Route>`/`<Redirect>` API to v6 `<Routes>`/`<Route>` via `react-router-dom-v5-compat`, following the pattern used for the earlier screen conversions.

- `Credentials.js` (list router): `<Switch>` → `<Routes>`; child routes use the `element` prop; the detail route is `/credentials/:id/*` so the nested `<Credential>` tree can match the rest of the path.
- `Credential.js` (detail router): `<Switch>` → `<Routes>` with **relative** child paths (`details`, `edit`, `access`, `job_templates`); the `exact` `<Redirect>` becomes an index route that `<Navigate replace>`s to `details`; the two duplicated v5 catch-all not-found routes collapse to a single `path="*"` route, and the `useRouteMatch({path})` `match.params.id` is replaced with the `id` route param from `useParams`.
- Tests: both suites rewritten with `renderWithContexts` (RTL) to assert **real v6 route resolution** — each tab panel, the kind-gated Job Templates tab, the index redirect, the unknown-sub-route not-found error, and the 404 detail error.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

- Tests: 12 route-resolution tests across the two converted suites; full `screens/Credential` directory passes (23 suites / 121 tests).

### DIFF
--- a/awx/ui/src/screens/Credential/Credential.js
+++ b/awx/ui/src/screens/Credential/Credential.js
@@ -3,15 +3,14 @@ import { useLingui } from '@lingui/react/macro';
 
 import { CaretLeftIcon } from '@patternfly/react-icons';
 import { Card, PageSection } from '@patternfly/react-core';
+import { Link } from 'react-router-dom';
 import {
-  Switch,
+  Routes,
+  Route,
+  Navigate,
   useParams,
   useLocation,
-  useRouteMatch,
-  Route,
-  Redirect,
-  Link,
-} from 'react-router-dom';
+} from 'react-router-dom-v5-compat';
 import useRequest from 'hooks/useRequest';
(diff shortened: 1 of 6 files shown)


blaipr:
Thanks for the review. Addressed in `f9b0e03690` and `a6824d4acb`: fixed the doubled `/* /*` comment and added back a ScreenHeader assertion (stream type and breadcrumb config) so the breadcrumb/title coverage the enzyme suite had is restored.

cigamit (approved):

PR #415, opened 2026-06-14
Convert Host route tree to react-router v6 Routes

blaipr:
##### SUMMARY

Continues the react-router v5 → v6 route-tree migration. Converts the **Host** screen's route trees — list, detail, and the nested `HostGroups` subtree — from the v5 `<Switch>`/`<Route>`/`<Redirect>` API to v6 `<Routes>`/`<Route>` via `react-router-dom-v5-compat`, following the pattern used for the earlier screen conversions.

- `Hosts.js` (list router): `<Switch>` → `<Routes>`; child routes use the `element` prop; the detail route is `/hosts/:id/*` so the nested `<Host>` tree can match the rest of the path.
- `Host.js` (detail router): `<Switch>` → `<Routes>` with **relative** child paths (`details`, `edit`, `facts`, `groups/*`, `jobs`); the `exact` `<Redirect>` becomes an index route that `<Navigate replace>`s to `details`; the catch-all not-found route stays as `path="*"`; `useRouteMatch` is replaced with the `id` route param, and the data routes stay guarded on the loaded `host`.
- `HostGroups/HostGroups.js` (nested subtree): its single absolute-path `<Switch>` route becomes an index `<Route>` under the `groups/*` parent.
- Tests: the three suites rewritten with `renderWithContexts` (RTL) to assert **real v6 route resolution** — each tab panel, the groups subtree, the index redirect, the unknown-sub-route not-found error, and the 404 detail error.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

- Like InstanceGroup, this converts a parent and its nested in-screen subtree (`HostGroups`) together so the relative `groups/*` delegation resolves under the v6 parent.
- Tests: 12 route-resolution tests across the three converted suites; full `screens/Host` directory passes (13 suites / 60 tests).

### DIFF
--- a/awx/ui/src/screens/Host/Host.js
+++ b/awx/ui/src/screens/Host/Host.js
@@ -2,14 +2,14 @@ import React, { useCallback, useEffect } from 'react';
 
 import { useLingui } from '@lingui/react/macro';
 
+import { Link } from 'react-router-dom';
 import {
-  Switch,
+  Routes,
   Route,
-  Redirect,
-  Link,
-  useRouteMatch,
+  Navigate,
+  useParams,
   useLocation,
-} from 'react-router-dom';
+} from 'react-router-dom-v5-compat';
 import { CaretLeftIcon } from '@patternfly/react-icons';
 import { Card, PageSection } from '@patternfly/react-core';
 import RoutedTabs from 'components/RoutedTabs';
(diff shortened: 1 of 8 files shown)


blaipr:
Thanks for the review. Addressed in `093d8bab07`: added a `path="*"` not-found fallback to the HostGroups routes so unknown sub-paths under `groups/` render a ContentError rather than a blank panel.

cigamit (approved):

PR #416, opened 2026-06-14
Convert WorkflowApproval route tree to react-router v6 Routes

blaipr:
##### SUMMARY

Continues the react-router v5 → v6 route-tree migration. Converts the **WorkflowApproval** screen's route trees from the v5 `<Switch>`/`<Route>`/`<Redirect>` API to v6 `<Routes>`/`<Route>` via `react-router-dom-v5-compat`, following the pattern used for the earlier screen conversions.

- `WorkflowApprovals.js` (list router): `<Switch>` -> `<Routes>`; child routes use the `element` prop; the detail route is `/workflow_approvals/:id/*` so the nested `<WorkflowApproval>` tree can match the rest of the path; the unused `useRouteMatch` `match.url` is replaced by explicit paths.
- `WorkflowApproval.js` (detail router): `<Switch>` -> `<Routes>` with a relative `details` child path; the `exact` `<Redirect>` becomes an index route that `<Navigate replace>`s to `details`; the catch-all not-found route stays as `path="*"`; `match.url` / `match.params.id` are replaced with the `workflowApprovalId` route param.
- Tests: both suites rewritten with `renderWithContexts` (RTL) to assert real v6 route resolution - the detail panel, the index redirect, the unknown-sub-route not-found error, and the 404 detail error.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

- Tests: 6 route-resolution tests across the two converted suites; full `screens/WorkflowApproval` directory passes (11 suites / 61 tests).

### DIFF
--- a/awx/ui/src/screens/WorkflowApproval/WorkflowApproval.js
+++ b/awx/ui/src/screens/WorkflowApproval/WorkflowApproval.js
@@ -3,15 +3,14 @@ import { useLingui } from '@lingui/react/macro';
 
 import { Card, PageSection } from '@patternfly/react-core';
 import { CaretLeftIcon } from '@patternfly/react-icons';
+import { Link } from 'react-router-dom';
 import {
-  Link,
-  Switch,
+  Routes,
   Route,
-  Redirect,
+  Navigate,
   useParams,
-  useRouteMatch,
   useLocation,
-} from 'react-router-dom';
+} from 'react-router-dom-v5-compat';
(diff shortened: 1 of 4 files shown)


blaipr:
Thanks for the review. Fixed the doubled `/* /*` in the nested-route comment in `1c6ed7956f`.

cigamit (approved):

PR #417, opened 2026-06-14
Convert Job route tree to react-router v6 Routes

blaipr:
##### SUMMARY

Continues the react-router v5 → v6 route-tree migration. Converts the **Job** screen's route trees - including the job-type disambiguation redirects - from the v5 `<Switch>`/`<Route>`/`<Redirect>` API to v6 `<Routes>`/`<Route>` via `react-router-dom-v5-compat`.

- `Jobs.js` (list/redirect router): `<Switch>` -> `<Routes>` with explicit `/jobs` paths (the `useRouteMatch` `match.path` is gone); the typed detail route is `/jobs/:typeSegment/:id/*` so the nested `<Job>` tree matches `details`/`output`; the untyped `/jobs/:id` and `/jobs/:id/(details|output)` routes still funnel through `<JobTypeRedirect>`; the legacy `/jobs/system/:id` redirect becomes a dedicated `/jobs/system/:id/*` route that `<Navigate>`s to `/jobs/management/:id`, preserving any sub-path.
- `JobTypeRedirect.js`: the standalone `<Redirect from={path} to>` becomes `<Navigate to replace>`, which drops the `path` / `useRouteMatch` dependency entirely.
- `Job.js` (detail router): `<Switch>` -> `<Routes>` with relative child paths (`details`, `output`); the exact index `<Redirect>` becomes an index route that `<Navigate>`s to `output`; `isWorkflow` is derived from `typeSegment` instead of `match.url`; tab links are built from the `typeSegment`/`id` route params.
- Tests: `Jobs.test.js` rewritten with `renderWithContexts` (RTL) to assert real v6 route resolution (list, typed detail subtree, untyped type redirects, and the system->management redirect); `Job.test.js` useParams mock pointed at `react-router-dom-v5-compat`.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

- v6 route ranking keeps the static `/jobs/:id/details`, `/jobs/:id/output` and `/jobs/system/...` routes ahead of the all-dynamic `/jobs/:typeSegment/:id/*` splat, preserving the original v5 `<Switch>` order.
- Tests: full `screens/Job` directory passes (19 suites / 157 tests). `npm --prefix awx/ui run lint` clean on the changed source.

### DIFF
--- a/awx/ui/src/screens/Job/Job.js
+++ b/awx/ui/src/screens/Job/Job.js
@@ -1,12 +1,11 @@
 import React, { useEffect, useCallback, useRef } from 'react';
+import { Link } from 'react-router-dom';
 import {
+  Routes,
   Route,
-  Switch,
-  Redirect,
-  Link,
+  Navigate,
   useParams,
-  useRouteMatch,
-} from 'react-router-dom';
+} from 'react-router-dom-v5-compat';
 import { useLingui } from '@lingui/react/macro';
 import { CaretLeftIcon } from '@patternfly/react-icons';
 import { Card, PageSection } from '@patternfly/react-core';
@@ -37,7 +36,6 @@ export const JOB_URL_SEGMENT_MAP = {
 function Job({ setBreadcrumb }) {
   const { t } = useLingui();
   const { id, typeSegment } = useParams();
-  const match = useRouteMatch();
 
   const type = JOB_URL_SEGMENT_MAP[typeSegment];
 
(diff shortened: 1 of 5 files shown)


blaipr:
Thanks for the review. Addressed in `62f6dcd7a2`: the JobTypeRedirect mock now mirrors the real `view=output` default and the untyped `/jobs/:id` test asserts `output` instead of a loose regex; and added a test that the legacy `/jobs/system/:id/*` redirect preserves the trailing sub-path (`.../system/5/output` to `.../management/5/output`).

cigamit (approved):

PR #418, opened 2026-06-14
Convert User route tree to react-router v6 Routes

blaipr:
##### SUMMARY

Continues the react-router v5 → v6 route-tree migration. Converts the **User** screen's nested route trees - the Users list, the User detail, the UserTokens subtree, and the UserToken detail - from the v5 `<Switch>`/`<Route>`/`<Redirect>` API to v6 `<Routes>`/`<Route>` via `react-router-dom-v5-compat`.

- `Users.js` (list router): `<Switch>` -> `<Routes>` with explicit `/users` paths; the detail route is `/users/:id/*` so the nested `<User>` tree matches; the unused `useRouteMatch` `match.path` is dropped.
- `User.js` (detail router): `<Switch>` -> `<Routes>` with relative child paths (`details`, `edit`, `organizations`, `teams`, `roles`, `tokens/*`); the exact `<Redirect>` becomes an index route that `<Navigate>`s to `details`; the catch-all not-found route stays as `path="*"`; `match.url` / `match.params.id` are replaced with the `id` route param.
- `UserTokens.js` (nested subtree): `<Switch>` -> `<Routes>` with relative paths (`add`, `:tokenId/*`, index list).
- `UserToken.js` (token detail): `<Switch>` -> `<Routes>` with a relative `details` path and an index `<Navigate>` to `details`.
- Descendants that read params under the now-v6 tree are switched to read from `react-router-dom-v5-compat` (UserTeamList, UserOrganizationList, UserTokenList, UserTokenListItem, UserTokenAdd, UserTokenDetail); UserList builds its add/detail links from the literal `/users` base instead of `useRouteMatch` `match.url`.
- Tests: `useParams` mocks pointed at `react-router-dom-v5-compat`, and components mounted under their real v6 parent route so the nested `<Routes>` resolve.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

- This is the deepest nesting converted so far (Users -> User -> UserTokens -> UserToken), each level delegating with a trailing `/*` and using relative child paths.
- Includes the same descendant param-resolution fixes that the earlier converted screens needed (the v5 `useParams`/`useRouteMatch` would otherwise return empty under a v6 parent).
- Tests: full `screens/User` directory passes (22 suites / 94 tests). `npm --prefix awx/ui run lint` clean on the changed source.

### DIFF
--- a/awx/ui/src/screens/User/User.js
+++ b/awx/ui/src/screens/User/User.js
@@ -1,14 +1,14 @@
 import React, { useEffect, useCallback } from 'react';
 
 import { useLingui } from '@lingui/react/macro';
+import { Link } from 'react-router-dom';
 import {
-  Switch,
+  Routes,
   Route,
-  Redirect,
-  Link,
-  useRouteMatch,
+  Navigate,
+  useParams,
   useLocation,
-} from 'react-router-dom';
+} from 'react-router-dom-v5-compat';
 import { CaretLeftIcon } from '@patternfly/react-icons';
 import { Card, PageSection } from '@patternfly/react-core';
 import useRequest from 'hooks/useRequest';
@@ -25,7 +25,7 @@ import UserRolesList from './UserRoles/UserRolesList';
 function User({ setBreadcrumb, me }) {
   const { t } = useLingui();
   const location = useLocation();
-  const match = useRouteMatch('/users/:id');
+  const { id } = useParams();
   const userListUrl = `/users`;
   const {
     result: user,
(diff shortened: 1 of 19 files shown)


blaipr:
Thanks for the review. Addressed in `f79aaea38d`: dropped the redundant `id` props (UserTokenAdd, UserToken and UserTokenList all read `id` from `useParams`), removed the now-unused import in UserTokens, and fixed the doubled `/* /*` comment.

cigamit (approved):

PR #419, opened 2026-06-14
Convert AllSchedules route tree to react-router v6 Routes

blaipr:
##### SUMMARY

Continues the react-router v5 → v6 route-tree migration. Converts the global **Schedules** screen (`AllSchedules`) from the v5 `<Switch>` API to v6 `<Routes>` via `react-router-dom-v5-compat`.

UI (no behavior change - same URL, same list):

- `AllSchedules.js`: `<Switch>` -> `<Routes>` with the `/schedules` list route using the `element` prop. `AllSchedules` renders the shared `ScheduleList` directly (not the nested `Schedules` router), so this is a single list route with no entanglement.
- Test rewritten with `renderWithContexts` (RTL) to mount at `/schedules` and assert the list resolves and the breadcrumb config is set.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

- `npm --prefix awx/ui run lint` clean; `screens/Schedule/AllSchedules` test passes.

### DIFF
--- a/awx/ui/src/screens/Schedule/AllSchedules.js
+++ b/awx/ui/src/screens/Schedule/AllSchedules.js
@@ -1,5 +1,5 @@
 import React, { useCallback } from 'react';
-import { Routes, Route } from 'react-router-dom-v5-compat';
+import { Routes, Route, Navigate } from 'react-router-dom-v5-compat';
 import { PageSection, Card } from '@patternfly/react-core';
 import { useLingui } from '@lingui/react/macro';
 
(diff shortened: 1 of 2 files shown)


blaipr:
Thanks for the review. Addressed in `a3ae435d11`: added a catch-all that redirects stray `/schedules/*` sub-paths back to the list (the v6 `<Routes>` only matched `/schedules` exactly), and guarded the ScreenHeader assertion with `toBeDefined()` so a missing render fails clearly.

cigamit (approved):

PR #420, opened 2026-06-14
Convert the Setting screen route trees to react-router v6 Routes

blaipr:
##### SUMMARY

Continues the react-router v5 → v6 route-tree migration. Converts the **entire Setting screen** - all 16 route trees - from the v5 `<Switch>`/`<Route>`/`<Redirect>` API to v6 `<Routes>`/`<Route>` via `react-router-dom-v5-compat`.

- **Single-form sub-screens** (TACACS, RADIUS, OIDC, SAML, GoogleOAuth2, Jobs, UI, Troubleshooting, MiscSystem, MiscAuthentication, Logging, Subscription): `<Switch>` → `<Routes>`; details/edit use the `element` prop; the base `<Redirect>` becomes a `<Route path={baseURL}>` that `<Navigate>`s to details; not-found kept as `{baseURL}/*`. MiscSystem/MiscAuthentication/Logging keep their superuser-gated edit route (the fallback inner `<Redirect>` becomes `<Navigate>`); Subscription folds its `useRouteMatch` base-redirect into a `<Route path={baseURL}>`.
- **Category-based sub-screens** (LDAP, AzureAD, GitHub): the `useRouteMatch` base/category redirects become a `<Route path={baseURL}>` Navigate plus a small `CategoryRedirect` helper (`useParams`) on the `/:category` route; detail/edit routes use the `element` prop.
- **`Settings.js` dispatcher**: `<Switch>` → `<Routes>`; each sub-screen route gets a trailing `/*` so its own nested `<Routes>` resolves; the non-superuser `<Redirect to="/">` and the open-license subscription `<Redirect>` become `<Navigate>`.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

- Part of the incremental react-router v6 migration; each screen's route tree is independent and converted on its own branch off `main`. This PR covers the whole Setting screen.
- The sub-screens keep their absolute `baseURL` paths. Under the now-v6 `Settings.js` dispatcher they nest as `/settings/<name>/*` → the sub-screen's own `<Routes>`; v6 permits those absolute child paths because they extend the parent `/settings/<name>` path.
- The existing enzyme tests mount sub-screens at absolute `/settings/<name>/...` URLs, which v6 `<Routes>` resolve unchanged - so the only test change is `Settings.test.js` asserting the non-superuser redirect via the history location instead of finding a `<Redirect>` component.
- Tests: full `screens/Setting` directory passes (61 suites / 344 tests). `npm --prefix awx/ui run lint` clean on the changed source.

### DIFF
--- a/awx/ui/src/screens/Setting/AzureAD/AzureAD.js
+++ b/awx/ui/src/screens/Setting/AzureAD/AzureAD.js
@@ -28,24 +28,24 @@ function AzureAD() {
       <Card>
         <Routes>
           <Route
-            path={baseURL}
+            index
             element={<Navigate to={`${baseURL}/default/details`} replace />}
           />
           <Route
-            path={`${baseURL}/:category`}
+            path=":category"
             element={<CategoryRedirect baseURL={baseURL} />}
           />
           <Route
-            path={`${baseURL}/:category/details`}
+            path=":category/details"
             element={<AzureADDetail />}
           />
-          <Route path={`${baseURL}/default/edit`} element={<AzureADEdit />} />
+          <Route path="default/edit" element={<AzureADEdit />} />
           <Route
-            path={`${baseURL}/tenant/edit`}
+            path="tenant/edit"
             element={<AzureADTenantEdit />}
           />
           <Route
-            path={`${baseURL}/*`}
+            path="*"
             element={
               <ContentError isNotFound>
                 <Link to={`${baseURL}/default/details`}>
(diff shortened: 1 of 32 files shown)


blaipr:
Thanks for the review. Renamed the redirect test in `b3ec2c3fd4`: it asserts a redirect via history after a `<Navigate>`, not a v5 `<Redirect>` component, so the name now matches the assertion.

cigamit:
There seems to be an issue with this one.  For some reason all settings pages are no longer displaying the actual settings.  Verified not a caching issue.  It is no longer making the API call to pull the setting for that page.  For example, clicking on the Logging Settings no longer creates a query for /api/v2/settings/logging/

image

cigamit (approved):
Settings pages now properly display

PR #421, opened 2026-06-14
Convert the Inventory screen route trees to react-router v6 Routes

blaipr:
##### SUMMARY

Continues the react-router v5 → v6 route-tree migration. Converts the **Inventory screen** - the largest remaining screen family (16 nested route trees) - from the v5 `<Switch>`/`<Route>`/`<Redirect>` API to v6 `<Routes>`/`<Route>` via `react-router-dom-v5-compat`. Done as a single PR with several commits.

UI (no behavior change - same URLs, same panels). Each route tree keeps its absolute `/inventories/...` paths, so it resolves correctly regardless of conversion order; every route that delegates to a nested screen gets a trailing `/*`; `useRouteMatch` is replaced with `useParams` + an explicit base url; exact `<Redirect>`s (and the inventory-kind mismatch redirects) become `<Navigate>`.

Converted (15 of the 16 trees):

- **Hosts**: AdvancedInventoryHost(s), InventoryHosts, InventoryHost, InventoryHostGroups
- **Groups**: InventoryGroups, InventoryGroup, InventoryGroupHosts, InventoryRelatedGroups
- **Sources**: InventorySources (the list)
- **Inventory-type details**: Inventory, SmartInventory, ConstructedInventory, FederatedInventory
- **Dispatcher**: Inventories.js (delegates each inventory-type detail with `/*`)

Deferred (1 tree):

- **InventorySource** (the source detail) renders the shared `Schedules` component, so it stays on v5 here and is handled with the shared-`Schedules` cluster. Its v5 `<Switch>` keeps working under the now-v6 `InventorySources` parent because it matches the absolute location.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

- Part of the incremental react-router v6 migration; converting the Inventory screen as one PR rather than many small PRs.
- Tests: full `screens/Inventory` directory passes (76 suites / 345 tests). A couple of tests needed updates (drop conflicting hoisted `useParams` module mocks and mount under the real v6 route; stub the inventory list / mount at `/inventories`); the rest pass unchanged because the routes keep absolute paths. `npm --prefix awx/ui run lint` clean on the changed source.

### DIFF
--- a/awx/ui/src/components/Schedule/Schedule.js
+++ b/awx/ui/src/components/Schedule/Schedule.js
@@ -1,14 +1,14 @@
 import React, { useEffect, useCallback } from 'react';
 import { useLingui } from '@lingui/react/macro';
 
+import { Link } from 'react-router-dom';
 import {
-  Switch,
+  Routes,
   Route,
-  Link,
-  Redirect,
+  Navigate,
   useLocation,
   useParams,
-} from 'react-router-dom';
+} from 'react-router-dom-v5-compat';
 import { CaretLeftIcon } from '@patternfly/react-icons';
 import { SchedulesAPI } from 'api';
 import useRequest from 'hooks/useRequest';
@@ -31,7 +31,7 @@ function Schedule({
 
   const { pathname } = useLocation();
 
-  const pathRoot = pathname.substr(0, pathname.indexOf('schedules'));
+  const pathRoot = pathname.substring(0, pathname.indexOf('schedules'));
 
   const {
     isLoading,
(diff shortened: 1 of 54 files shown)


blaipr:
Thanks for the review. Fixed in `b5d5370139`: the not-found link built a literal `/inventories/:inventoryType/...` segment; it now interpolates `inventoryType` from `useParams` so the link resolves.

cigamit:
There is an issue with this one.  Clicking on an inventory, all tabs (Details, Access, Groups, etc..) throw the error
"The page you requested could not be found."

image

cigamit:
Details and a few other pages now work, but still getting errors when clicking on Groups, Hosts, or Sources in an inventory

Page not found at /api/v2/inventories/undefined/ad_hoc_commands/
Page not found at /api/v2/inventories/undefined/hosts/
Page not found at /api/v2/inventories/undefined/inventory_sources/
Page not found at /api/v2/inventories/undefined/groups/

image

blaipr:
Pushed a fix for the Groups/Hosts/Sources errors you hit (the /api/v2/inventories/undefined/... calls).

Root cause: the inventory route tree is v6 now, but its leaf components (the lists, detail, add/edit and list-item components) were still reading :id / :inventoryType / :groupId through react-router v5 useParams, which returns nothing inside a v6 Routes tree. So the id came through as undefined and the API calls 404'd.

Changes:
- Read those params from react-router-dom-v5-compat in the 16 affected leaf components.
- Also pulled in the v6 shared Schedule component so the inventory source Schedules tab resolves.
- Reworked the tests to mount each component under a real v6 route instead of mocking the router, so they exercise the actual routing.

Verified in a local browser against real data: Details, Access, Groups, Hosts, Sources, the inventory list, and the inventory source Schedules list and detail all load correctly now.

cigamit (approved):

PR #422, opened 2026-06-14
Convert the shared Schedule components to react-router v6 Routes

blaipr:
##### SUMMARY

Continues the react-router v5 → v6 route-tree migration. Converts the **shared Schedule components** (`components/Schedule/Schedules.js` and `Schedule.js`) from the v5 `<Switch>`/`<Redirect>` API to v6 `<Routes>`/`<Navigate>` via `react-router-dom-v5-compat`.

These components are reused at several different base paths (templates, projects, inventory sources, management jobs), so they can't hardcode an absolute base. `Schedule.js` already derived its base from the location (`pathRoot` = the pathname up to `schedules`); this applies the same technique to `Schedules.js`.

- `Schedules.js`: drop `useRouteMatch`; compute the base from `useLocation` and use it for the list / add / `:scheduleId` routes (the schedule detail delegates with `/*`).
- `Schedule.js`: `<Switch>` → `<Routes>`; the exact `<Redirect>` becomes a `<Route>` that `<Navigate>`s to details; the edit/details routes keep their computed `pathRoot` paths; not-found stays `path="*"`.

Because the routes are computed-absolute, they resolve whether the parent screen is still on v5 `<Switch>` or already migrated to v6. This **decouples** the shared-component conversion from the screens that mount `<Schedules>` (Template, Project, ManagementJob, InventorySource, WorkflowJobTemplate), which can now be migrated independently afterward.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

- Part of the incremental react-router v6 migration. This unblocks the last cluster of screens (the five that embed the schedule subtree).
- Tests: `components/Schedule` passes (13 suites / 126 tests), and all five parent screens that mount `<Schedules>` pass unchanged. `npm --prefix awx/ui run lint` clean on the changed source.

### DIFF
--- a/awx/ui/src/components/Schedule/Schedule.js
+++ b/awx/ui/src/components/Schedule/Schedule.js
@@ -1,14 +1,8 @@
 import React, { useEffect, useCallback } from 'react';
 import { useLingui } from '@lingui/react/macro';
 
-import {
-  Switch,
-  Route,
-  Link,
-  Redirect,
-  useLocation,
-  useParams,
-} from 'react-router-dom';
+import { Link, useLocation, useParams } from 'react-router-dom';
+import { Routes, Route, Navigate } from 'react-router-dom-v5-compat';
 import { CaretLeftIcon } from '@patternfly/react-icons';
 import { SchedulesAPI } from 'api';
 import useRequest from 'hooks/useRequest';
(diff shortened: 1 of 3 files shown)


blaipr:
Thanks for the review. Addressed in `4b0b60942f`: Schedule.js now imports `useParams`/`useLocation` from `react-router-dom-v5-compat` so `scheduleId` resolves under the v6 parent, and the edit route uses `:scheduleId` consistently; Schedules.js derives `baseUrl` with `substring` and handles the case where `schedules` is absent so it never collapses to a relative path; and the doubled `/* /*` comment is fixed.

cigamit:
After creating a Job Template or Workflow schedule, it fails when showing the details.

image

cigamit:
Regressed a bit on this one, now can not list schedules of a Job Template
Page not found at /api/v2/schedules/templates/
image

blaipr:
About the 'can not list schedules of a Job Template / Page not found at /api/v2/schedules/templates/' regression.

This PR converts only the shared Schedule component to v6 relative routes. The catch is that the relative routes only resolve when the Schedule component is mounted under a v6 consumer screen (Template, Project, ManagementJob, Inventory source). On this branch the consumer screens are still the v5 versions from main, so the relative routes resolve against the wrong base and you get /api/v2/schedules/<segment>/.

In other words the shared Schedule conversion and the consumer screen conversions have to land together. I have cherry-picked this exact v6 Schedule change into each consumer PR (#421, #423, #424, #427), so it is already carried there and verified working in a browser in that context.

Suggestion: merge this one first and the consumer PRs on top (their cherry-picks become no-ops), or fold it into the consumers and close this. Either works, it just should not merge or be tested in isolation. Happy to do whichever you prefer.

blaipr:
Closing this as redundant. The shared v6 Schedule conversion already landed on main through #424, which carried the same two-file change as a cherry-pick.

Verified against the current main: this branch's only contribution is Schedule.js, Schedules.js and Schedules.test.js, and all three are now byte-identical to main (0 lines of difference). Merging this PR would apply nothing.

The four consumer screens that delegate to Schedules each carry the same v6 Schedule change so they are self-contained and verified working in a browser: Inventory #421, ManagementJob #423, Project #424 (merged), Template #427. InventorySource #426 is superseded by #421.

Nothing is lost by closing this.

PR #423, opened 2026-06-14
Convert ManagementJob route tree to react-router v6 Routes

blaipr:
##### SUMMARY

Continues the react-router v5 → v6 route-tree migration. Converts the **ManagementJob** screen's route trees from the v5 `<Switch>`/`<Route>`/`<Redirect>` API to v6 `<Routes>`/`<Route>` via `react-router-dom-v5-compat`.

- `ManagementJobs.js` (list router): `<Switch>` → `<Routes>`; the detail route is `/management_jobs/:id/*` so the nested `<ManagementJob>` tree matches.
- `ManagementJob.js` (detail router): `<Switch>` → `<Routes>`; `useRouteMatch` is replaced with the `id` route param + an explicit `detailUrl`; the exact `<Redirect>` becomes a `<Route>` that `<Navigate>`s to `schedules` (the default tab); the schedules tab delegates with `schedules/*` to the shared `<Schedules>`; notifications uses the `element` prop.
- `ManagementJobs.test.js`: mount at `/management_jobs` with a stubbed list so the route resolves without an API call.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

- Part of the incremental react-router v6 migration. One of the screens that embeds the shared schedule subtree.
- **Depends on #422** (shared Schedule component conversion): converting this screen to v6 while `Schedules` is still v5 would reintroduce the `match.path` break, so #422's two component files are included in this branch until it merges (then this rebases onto main and they drop out).
- Tests: full `screens/ManagementJob` directory passes (3 suites / 7 tests). `npm --prefix awx/ui run lint` clean on the changed source.

### DIFF
--- a/awx/ui/src/components/Schedule/Schedule.js
+++ b/awx/ui/src/components/Schedule/Schedule.js
@@ -1,14 +1,8 @@
 import React, { useEffect, useCallback } from 'react';
 import { useLingui } from '@lingui/react/macro';
 
-import {
-  Switch,
-  Route,
-  Link,
-  Redirect,
-  useLocation,
-  useParams,
-} from 'react-router-dom';
+import { Link, useLocation, useParams } from 'react-router-dom';
+import { Routes, Route, Navigate } from 'react-router-dom-v5-compat';
 import { CaretLeftIcon } from '@patternfly/react-icons';
 import { SchedulesAPI } from 'api';
 import useRequest from 'hooks/useRequest';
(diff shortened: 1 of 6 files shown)


blaipr:
Thanks for the review. Addressed in `038450f83d`: the shared Schedules `baseUrl` now uses `substring` and handles the `schedules`-absent case explicitly (carried from the #422 fix this stacks on).

cigamit:
Schedules fail to display details when clicked
image

cigamit:
Management jobs are listed under schedules, but when clicking, its just blank data
image

blaipr:
Fixed the blank schedule detail.

ManagementJob is rendered as a descendant of the v6 ManagementJobs dispatcher (path management_jobs/:id/*), so its nested Routes resolves relative to management_jobs/:id. The child routes were using absolute paths (management_jobs/:id/schedules/*), which do not match in a descendant route tree, so clicking a schedule never mounted the Schedule detail (and never fetched it). Switched them to relative paths (index, notifications, schedules/*).

Verified in a local browser: the schedule list renders and clicking a schedule now loads /api/v2/schedules/<id>/ and shows the full detail.

cigamit (approved):

PR #425, opened 2026-06-14
Convert the App.js root route tree to react-router v6 Routes

blaipr:
##### SUMMARY

Converts the **App.js root route tree** to react-router v6 `<Routes>`. This is the last piece of the v5 to v6 migration (#400 through #427). App.js now dispatches each `routeConfig` screen as a v6 descendant at `path="/<x>/*"`, and the outer login, `/` to `/home`, and `ProtectedRoute` tree is plain v6.

Once App.js is v6, every screen's own `<Routes>` becomes a descendant, which v6 matches relative to the parent. The merged route-tree PRs had written each list-entry screen with absolute child paths (`path="/credentials"`, `/credentials/add`, `/credentials/:id/*`), and those only matched while App.js was still v5 and the screens were effectively top-level. As descendants they stop matching, so most pages went blank or not-found after rebasing onto current `main`. This PR switches those screens to relative paths so they resolve under the v6 root:

- List-entry screens use `add`, `:id/*`, and `<Route index>`: Credentials, Jobs, Projects, Inventories, Hosts, Organizations, Users, Teams, CredentialTypes, WorkflowApprovals, NotificationTemplates, Applications, ExecutionEnvironments, AllSchedules, InstanceGroups, Instances, Settings, ManagementJobs.
- Template and WorkflowJobTemplate detail `<Routes>` use relative `:templateType/:id/...` (Templates.js itself stays a v5-compat `<Switch>`, which already prefix-matches the full path).
- Each screen's tests now mount the screen as a descendant under a real `<Routes><Route path="/<x>/*" element={<Screen/>}/></Routes>`. The old top-level mounts only resolved with absolute paths.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

I browser-smoke-tested this against the dev server (hash routing) after the change. Every top-level route and a representative set of detail tabs render, with no `/api/v2/.../undefined/...` calls:

- Lists: home, jobs, schedules, activity_stream, workflow_approvals, labels, templates, credentials, projects, inventories, hosts, organizations, users, teams, credential_types, notification_templates, management_jobs, applications, execution_environments, instances, topology_view, settings. All OK.
- Details: credentials/:id, jobs/:id/output, projects/:id, inventories/inventory/:id, hosts/:id, organizations/:id, instance_groups/:id, management_jobs/:id (notifications and schedules), settings/<category>, and templates/{job_template,workflow_job_template}/:id (details/edit/survey/visualizer). All OK.
- `host_metrics` and `subscription_usage` are license-gated out of `routeConfig` (`SUBSCRIPTION_USAGE_MODEL`), so they 404 by design. That's unchanged from `main`.

All 20 affected screen test suites pass (69 tests), and the changed screens lint clean.

### DIFF
--- a/awx/ui/src/App.js
+++ b/awx/ui/src/App.js
@@ -196,7 +196,7 @@ function App() {
   const redirectURL = window.sessionStorage.getItem(SESSION_REDIRECT_URL);
   if (redirectURL) {
     window.sessionStorage.removeItem(SESSION_REDIRECT_URL);
-    if (redirectURL !== '/' || redirectURL !== '/home')
+    if (redirectURL !== '/' && redirectURL !== '/home')
       navigate(redirectURL, { replace: true });
   }
 
--- a/awx/ui/src/App.js
+++ b/awx/ui/src/App.js
@@ -1,14 +1,13 @@
 import React, { useEffect, useLayoutEffect, useState } from 'react';
+import { HashRouter } from 'react-router-dom';
 import {
-  useRouteMatch,
-  useLocation,
-  HashRouter,
+  CompatRouter,
+  Routes,
   Route,
-  Switch,
-  Redirect,
-  useHistory,
-} from 'react-router-dom';
-import { CompatRouter } from 'react-router-dom-v5-compat';
+  Navigate,
+  useLocation,
+  useNavigate,
+} from 'react-router-dom-v5-compat';
 import { ErrorBoundary } from 'react-error-boundary';
 import locationReplace from 'util/navigation';
 import { I18nProvider } from '@lingui/react';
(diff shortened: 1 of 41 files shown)


blaipr:
Thanks for the review. Fixed in `172701386c`: `redirectURL !== / || redirectURL !== /home` was always true, so changed it to `&&` so the post-login redirect only runs when the stored URL is neither default.

blaipr:
Smoke-tested this in a browser against a running instance (login + 14 top-level routes). The root conversion works: login and the post-login redirect resolve, and the App.js routeConfig dispatcher routes correctly to Dashboard, Jobs, Templates, Projects, Inventories, Credentials, Organizations, Users, Teams, Schedules, Settings, Management Jobs, Notification Templates and Instance Groups. No errors, no not-found, no changes needed here.

blaipr:
On the `App.js` redirect condition Copilot flagged: this is a **false positive**. The current line is `if (redirectURL !== '/' && redirectURL !== '/home')`, which is only true when `redirectURL` is **neither** `/` nor `/home` - it correctly *skips* navigation for those two. Copilot's "always true" reasoning applies to the `||` form; in fact this PR **fixes** the pre-existing `||` bug (which *was* always true) by switching it to `&&`. No change needed here.

The rest of the v6 conversion checks out: `Routes` (login / `/`→`/home` / `*`→ProtectedRoute), `ProtectedRoute` as an element wrapper, `Navigate` replacements, and `/*` on each screen route for nested matching. Tests pass.

cigamit:
Now that all the other page changes are in, and I rebase this one to main, most of those other pages fail to load completely.  Things like dashboard and labels load, but the others are blank pages, no errors.

Also FYI, in main, the Templates > Schedules no longer loads, was gonna create an issue for it, but figured I could drop it here while you are working on this one.

blaipr:
@cigamit good catch, reproduced it.

The blank pages come from making `App.js` v6: each screen now mounts as a v6 descendant, but the merged route-tree PRs left the list-entry screens using **absolute** child paths (`/credentials`, `/credentials/add`, etc). Those worked while `App.js` was still v5 (the screens were effectively top-level), but as v6 descendants they no longer match, so the content comes up blank. Dashboard and Labels survived because they have no nested `<Routes>`, and Templates survived because it's still a `<Switch>`.

I have just pushed a fix here: the 18 entry screens plus the Template/WorkflowJobTemplate detail routes now use relative paths, and their tests remount the screen as a descendant. `App.js` itself is unchanged from what you reviewed. I browser-checked every top-level route plus the detail tabs, and everything loads with no `undefined` API calls. 20 suites green.

And the Templates > Schedules one is **solved**. All load for me now. Shout if you still see it on your end.

cigamit (approved):

PR #424, opened 2026-06-14
Convert Project route tree to react-router v6 Routes

blaipr:
##### SUMMARY

Continues the react-router v5 → v6 route-tree migration. Converts the **Project** screen's route trees from the v5 `<Switch>`/`<Route>`/`<Redirect>` API to v6 `<Routes>`/`<Route>` via `react-router-dom-v5-compat`.

UI (no behavior change - same URLs, same panels):

- `Projects.js` (list router): `<Switch>` → `<Routes>`; the detail route is `/projects/:id/*` so the nested `<Project>` tree matches.
- `Project.js` (detail router): `<Switch>` → `<Routes>` with relative child paths (`details`, `edit`, `access`, `notifications`, `job_templates`, `schedules/*`); the exact `<Redirect>` becomes an index route that `<Navigate>`s to `details`; the schedules tab delegates with `schedules/*` to the shared `<Schedules>`; not-found stays `path="*"`.
- Tests: mount `Project` under the real `/projects/:id/*` route (dropping the `react-router-dom` `useRouteMatch`/`useParams` mocks), stub the list in `Projects.test.js`, and mock `RootAPI.readAssetVariables` for the brand-name hook the now-resolved detail route mounts.

##### ISSUE TYPE

- Bug, Docs Fix or other nominal change

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

- Part of the incremental react-router v6 migration. One of the screens that embeds the shared schedule subtree.
- **Depends on #422** (shared Schedule component conversion): its two component files are included in this branch until #422 merges (then this rebases onto main and they drop out).
- Tests: full `screens/Project` directory passes (11 suites / 66 tests). `npm --prefix awx/ui run lint` clean on the changed source.

PR #426, opened 2026-06-14
Convert InventorySource route tree to react-router v6

blaipr:
##### SUMMARY

Converts the **InventorySource** route tree from react-router v5 to v6, continuing the react-router v6 migration (`react-router-dom-v5-compat` bridge).

- `<Switch>`/`<Route>`/`<Redirect>` -> `<Routes>`/`<Route>`/`<Navigate>`.
- Replaces `useRouteMatch('/inventories/inventory/:id/sources/:sourceId')` with `useParams`, building the absolute base path from the route params.
- The exact `<Redirect>` to the details tab becomes an index `<Route>` + `<Navigate replace>`.
- The schedules tab delegates to the shared v6 `<Schedules>` component, mounted with a trailing `/*` so its nested route tree matches.
- The not-found case uses `path="*"`.
- Test updated to mock `useParams` instead of `useRouteMatch`.

This is the InventorySource screen that was deferred from the Inventory PR (#421) because it renders the shared `<Schedules>` component.

##### DEPENDENCY

Stacked on #422 (the v6 `<Schedules>`/`<Schedule>` conversion). The diff includes those two shared files; once #422 merges this will rebase cleanly to a single-file change.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

### DIFF
--- a/awx/ui/src/components/Schedule/Schedule.js
+++ b/awx/ui/src/components/Schedule/Schedule.js
@@ -1,14 +1,8 @@
 import React, { useEffect, useCallback } from 'react';
 import { useLingui } from '@lingui/react/macro';
 
-import {
-  Switch,
-  Route,
-  Link,
-  Redirect,
-  useLocation,
-  useParams,
-} from 'react-router-dom';
+import { Link, useLocation, useParams } from 'react-router-dom';
+import { Routes, Route, Navigate } from 'react-router-dom-v5-compat';
 import { CaretLeftIcon } from '@patternfly/react-icons';
 import { SchedulesAPI } from 'api';
 import useRequest from 'hooks/useRequest';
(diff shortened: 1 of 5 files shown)


blaipr:
Thanks for the review. Addressed in `4599793f40`: carried the shared Schedule fix (the `useParams`/`useLocation` import source, the `:scheduleId` edit-route param, and the `baseUrl` hardening), which includes the `/* /*` comment cleanup.

blaipr:
Heads up on overlap. This PR converts InventorySource.js to v6, but its parent inventory screens (Inventories.js, Inventory.js, InventorySources.js) are still v5 on main, so the v6 child can not resolve its routes on this branch by itself (the source tabs and the source Schedules tab break).

#421 already converts InventorySource.js to v6 plus all of its parent screens and the inventory leaf components, and it is verified working end to end in a browser (including the inventory source detail and Schedules). So this PR looks superseded by #421.

Suggestion: close this in favor of #421, or if you want inventory-source kept separate, it needs the parent inventory screens converted too (which is what #421 does). Let me know which way you want to go.

blaipr:
Closing this as superseded by #421.

Every file this PR touches is also converted in #421, and #421 does it more completely: it converts InventorySource.js plus its parent inventory screens (Inventories.js, Inventory.js, InventorySources.js) and the inventory leaf components, whereas this branch only converts InventorySource.js.

That parent coverage matters. On this branch the inventory parents are still v5, so the v6 InventorySource child can not resolve its routes here (the source tabs and the source Schedules tab break in isolation). This branch's InventorySource.js also still reads useParams from react-router-dom (v5), which returns nothing inside a v6 tree. #421 fixes both: it reads params from react-router-dom-v5-compat and converts the parents, and it is verified working end to end in a browser, including the inventory source detail and the Schedules list and detail.

The v6 shared Schedule change this branch also carried is already on main via #424, so nothing here is unique. The inventory-source v6 conversion ships via #421.

PR #427, opened 2026-06-14
Convert Template and WorkflowJobTemplate route trees to react-router v6

blaipr:
##### SUMMARY

Converts the **Template**, **WorkflowJobTemplate** and **TemplateSurvey** route trees from react-router v5 to v6, finishing the react-router v6 migration for the Templates screen (`react-router-dom-v5-compat` bridge). These are the last remaining `<Switch>` screens.

- `<Switch>`/`<Route>`/`<Redirect>` -> `<Routes>`/`<Route>`/`<Navigate>` in Template.js, WorkflowJobTemplate.js and TemplateSurvey.js.
- Template / WorkflowJobTemplate keep the `:templateType/:id` route patterns so the matched params reach descendants through the compat layer. The exact `<Redirect>` to the details tab becomes an index `<Route>` + `<Navigate replace>`.
- The schedules and survey tabs delegate to the shared v6 `<Schedules>` and `<TemplateSurvey>` components, mounted with a trailing `/*` so their nested route trees match.
- `TemplateSurvey` derives its base path from the location and reads the template type/id from the `template` prop, so it no longer depends on the parent router. Its own `<Switch>` becomes `<Routes>`.
- The Survey toolbar/add/edit helpers derive their base path from the location instead of `useRouteMatch` (which returns `/` under a v6 parent).
- `JobTemplateDetail` reads its route params from `react-router-dom-v5-compat` so the id resolves under the converted parent.
- The Template not-found route is no longer gated on the loading flag, avoiding a transient render where no route matches the current location.
- `Templates.js` (the list/add dispatcher) stays on v5, matching the pattern used for the other dispatchers (Projects.js, InventorySources.js).

##### DEPENDENCY

Stacked on #422 (the v6 `<Schedules>`/`<Schedule>` conversion). The diff includes those two shared files; once #422 merges this will rebase cleanly.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

### DIFF
--- a/awx/ui/src/components/Schedule/Schedule.js
+++ b/awx/ui/src/components/Schedule/Schedule.js
@@ -1,14 +1,8 @@
 import React, { useEffect, useCallback } from 'react';
 import { useLingui } from '@lingui/react/macro';
 
-import {
-  Switch,
-  Route,
-  Link,
-  Redirect,
-  useLocation,
-  useParams,
-} from 'react-router-dom';
+import { Link, useLocation, useParams } from 'react-router-dom';
+import { Routes, Route, Navigate } from 'react-router-dom-v5-compat';
 import { CaretLeftIcon } from '@patternfly/react-icons';
 import { SchedulesAPI } from 'api';
 import useRequest from 'hooks/useRequest';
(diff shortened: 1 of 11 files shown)


blaipr:
Thanks for the review. Addressed in `59c0f23593`: aligned the survey-delete test URL id with the template data id (mounted at `.../job_template/7/survey` to match `mockJobTemplateData.id`) so the route id and `template.id` agree and the intent is clear.

cigamit:
Everything is working except for surveys, the page is blank for it, no errors in the logs.
image

blaipr:
Fixed the blank survey tab (cbc625f). Root cause: `TemplateSurvey` is a v6 descendant (Template mounts it at `.../:id/survey/*`), but its `<Routes>` used **absolute** `surveyUrl` paths, which don't match in a descendant tree - so nothing rendered. Switched to relative child paths (`add`/`edit`) with an `index` route for the survey list.

Browser-verified on a live instance (puppeteer): the survey list renders the questions, and `/survey/add` renders the add form - both previously blank.

Also addressed the Copilot note: the test now mounts `TemplateSurvey` under a **real v6 `survey/*` route** (instead of a v5 `<Route>` with mocked params), so it actually exercises the descendant routing - the old mocked test passed *with* the bug. Route ids now match the template ids, asserted with concrete literals (7/15).

cigamit (approved):

PR #433, opened 2026-06-16
enzyme -> RTL: convert the CredentialType screen suites

blaipr:
##### SUMMARY

Continue the enzyme → React Testing Library migration (step 3 of the React modernization) by converting the `screens/CredentialType` suites. This follows the infrastructure in #385 and the first directory batch (`screens/User`) in #398.

- Converts all six enzyme suites in `screens/CredentialType` to `renderWithContexts`: `CredentialTypeListItem`, `CredentialTypeList`, `CredentialTypeDetails`, `CredentialTypeAdd`, `CredentialTypeEdit`, and the shared `CredentialTypeForm`.
- The container suites (Add/Edit) stub the shared form and drive its `onSubmit`/`onCancel`/`submitError` props — the same black-box treatment the enzyme tests used (the form has its own suite).
- Delete flows now interact for real. PF4's `Modal` aria-hides the rest of the tree under jsdom, so the confirm button is queried by label and clicked via `fireEvent` (a reusable pattern for the remaining directories).
- Drops one prop-inspection-only assertion (`ToolbarDeleteButton` `deleteDetailsRequests` count) that has no DOM-observable behavior.

No application code changes — test-only. `screens/CredentialType` no longer references enzyme.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ASCENDER VERSION

```
awx: 25.4.1.dev49+g0e26da4e2e
```

##### ADDITIONAL INFORMATION

- All 33 tests in `screens/CredentialType` pass (`npm test`).
- Test files are not linted (ignored by the eslint flat config), and there are no behavior changes, so no browser smoke-test is required for this test-only batch.
- Part of the step-3 directory-by-directory conversion; both enzyme and RTL helpers coexist until the last enzyme import is removed.

### DIFF
--- a/awx/ui/src/screens/CredentialType/CredentialType.test.js
+++ b/awx/ui/src/screens/CredentialType/CredentialType.test.js
@@ -58,7 +58,7 @@ describe('<CredentialType />', () => {
   test('fetches the credential type detail', async () => {
     renderAt('/credential_types/42/details');
     expect(await screen.findByText('CredentialTypeDetails')).toBeInTheDocument();
-    // real route params are strings (the old enzyme test mocked a number)
+    // real route params are strings (the previous test mocked a number)
     expect(CredentialTypesAPI.readDetail).toHaveBeenCalledWith('42');
   });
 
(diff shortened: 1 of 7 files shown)


cigamit:
Copilot had a few suggestions but otherwise this seems to test out just fine.  Can approve after you resolve or discard the suggestions above.

blaipr:
Addressed the Copilot suggestions:
- `CredentialTypeForm` value-update test now exercises **both** Name and Description and asserts the inputs exist before typing.
- `CredentialTypeListItem` passes `rowIndex` and asserts the selection checkbox by its accessible name (`Select row 0`).
- `CredentialTypeList` selects row checkboxes by accessible name instead of a positional `getAllByRole('checkbox')` index.

All 33 tests in `screens/CredentialType` still pass.

cigamit (approved):

PR #437, opened 2026-06-16
enzyme -> RTL: convert the Instances screen suites

blaipr:
##### SUMMARY

Converts the **Instances** screen's test suite from enzyme to React Testing Library, continuing the incremental enzyme → RTL migration (one screen directory per PR).

- `InstanceList` / `InstanceListItem` — row data, progress bar, expanded-row details, hop-node hiding, capacity-slider fork recompute
- `InstanceDetail` — detail fields, health-check button state + error modal, instance enable/disable toggle, capacity slider
- `InstanceAdd` / `InstanceEdit` — create/update API args + redirect, cancel navigation, submit-error branch (shared `InstanceForm` mocked)
- `InstanceForm` — input updates and submit payload (tooltip-labelled fields queried by id)

Interactions now go through accessible roles and real user events. Behaviour and assertions are preserved.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

`npm test` for the Instances directory: **9 suites, 43 tests, all passing.** ESLint clean. No production code changed — test-only.

One list-item error-path test was adapted rather than asserting a visible modal: the component renders that `AlertModal` with no `isOpen` prop, so PatternFly never mounts it to the DOM (enzyme matched the closed React element; RTL sees no DOM node). The identical error branch is still exercised and the rejected API call asserted, with a code comment explaining the component quirk.

### DIFF
--- a/awx/ui/src/screens/Instances/InstanceAdd/InstanceAdd.test.js
+++ b/awx/ui/src/screens/Instances/InstanceAdd/InstanceAdd.test.js
@@ -1,52 +1,77 @@
 import React from 'react';
-import { act } from 'react-dom/test-utils';
+import { screen, waitFor } from '@testing-library/react';
 import { createMemoryHistory } from 'history';
 import { InstancesAPI } from 'api';
-import {
-  mountWithContexts,
-  waitForElement,
-} from '../../../../testUtils/enzymeHelpers';
+import { renderWithContexts } from '../../../../testUtils/rtlContexts';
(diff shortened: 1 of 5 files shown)


blaipr:
Thanks for the review. The capacity test now asserts the slider value computed from the 0.1 step rather than a hard-coded 1.

cigamit (approved):

PR #439, opened 2026-06-16
enzyme -> RTL: convert the Metrics screen suites

blaipr:
##### SUMMARY

Converts the **Metrics** screen's test suite from enzyme to React Testing Library, continuing the incremental enzyme → RTL migration (one screen directory per PR).

- `LineChart` — asserts the `#chart` container renders and the help text lands in it
- `Metrics` — empty state, instance/metric Select interactions driving the exact `MetricsAPI.read` call, and exclusion of receptor/execution nodes from the instance list

Interactions now go through accessible roles (Selects scoped by their `ouiaId`) and real user events. Behaviour and assertions are preserved.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

`npm test` for the Metrics directory: **2 suites, 4 tests, all passing.** ESLint clean. No production code changed — test-only.

### DIFF
--- a/awx/ui/src/screens/Metrics/LineChart.test.js
+++ b/awx/ui/src/screens/Metrics/LineChart.test.js
@@ -1,37 +1,38 @@
 import React from 'react';
-import { act } from 'react-dom/test-utils';
 
-import { mountWithContexts } from '../../../testUtils/enzymeHelpers';
+import { renderWithContexts } from '../../../testUtils/rtlContexts';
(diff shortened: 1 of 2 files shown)


blaipr:
Thanks for the review. The option lookups now query from screen (the open listbox) instead of scoping to the select container, so the Popper-rendered options are found reliably.

cigamit (approved):

PR #441, opened 2026-06-16
enzyme -> RTL: convert the HostMetrics screen suites

blaipr:
##### SUMMARY

Converts the **HostMetrics** screen's test suite from enzyme to React Testing Library, continuing the incremental enzyme → RTL migration (one screen directory per PR).

`HostMetrics.test.js` migrated off `mountWithContexts`/enzyme onto `renderWithContexts`: list rendering and row presence after data load. Interactions go through accessible roles and real user events. Behaviour and assertions are preserved.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

`npm test` for the HostMetrics directory: **1 suite, 2 tests, all passing.** ESLint clean. No production code changed — test-only.

### DIFF
--- a/awx/ui/src/screens/HostMetrics/HostMetrics.test.js
+++ b/awx/ui/src/screens/HostMetrics/HostMetrics.test.js
@@ -1,10 +1,7 @@
 import React from 'react';
-import { act } from 'react-dom/test-utils';
+import { screen, waitFor } from '@testing-library/react';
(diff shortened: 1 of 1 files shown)


blaipr:
Thanks for the review. The ignored match and location props are gone, and the host row count is asserted via an accessible role query instead of querySelectorAll.

cigamit (approved):

PR #447, opened 2026-06-16
enzyme -> RTL: convert the Organization screen suites

blaipr:
##### SUMMARY

Converts the **Organization** screen's test suite from enzyme to React Testing Library, continuing the incremental enzyme → RTL migration (one screen directory per PR).

- `OrganizationList` / `OrganizationListItem` — load, selection, bulk delete + related-delete-request count, deletion error
- `OrganizationDetail` — detail fields, empty instance-groups/galaxy-credentials, related-delete-request count
- `OrganizationAdd` / `OrganizationEdit` — create/update with instance-group associate/disassociate, galaxy credentials, max-hosts (shared `OrganizationForm` mocked)
- `OrganizationForm` — fields, lookups (mocked), default galaxy credential, instance group add
- `OrganizationExecEnvList` + item, `OrganizationTeamList` + item — list idioms

Interactions now go through accessible roles and real user events. Behaviour and assertions are preserved.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

`npm test` for the Organization directory: **12 suites, 62 tests, all passing.** ESLint clean. No production code changed — test-only.

Two redundant tests were folded rather than duplicated: an Add-mode "AnsibleSelect does not render" assertion (it tested the real form's internal rendering, which is now mocked in the Add container suite and covered by the `OrganizationForm` suite) and two trivial "should mount successfully" duplicates subsumed by stronger row-render assertions.


blaipr:
Thanks for the review. Pushed a commit: OrganizationDetail renders under a real route so the Edit link resolves to /organizations/12/edit instead of undefined, the dead settleTooltips call on the still-open error modal is dropped, update.mockResolvedValue now takes a single value, and the misspelled OrgnizationTeamListItem test file is renamed to OrganizationTeamListItem.

cigamit (approved):

PR #451, opened 2026-06-16
enzyme -> RTL: convert the Job screen suites

blaipr:
##### SUMMARY

Converts the **Job** screen's test suite from enzyme to React Testing Library, continuing the incremental enzyme → RTL migration (one screen directory per PR). This is the largest screen directory; it included two files importing `enzyme` directly (`WorkflowOutputLink`, `useJobEvents`) which are also converted, so the directory is now fully enzyme-free.

Files migrated off enzyme onto `renderWithContexts` / RTL `render`:

- `Job` — tab routing under a real v6 route
- `JobDetail` — detail fields, relaunch/cancel/delete + related-resource reads
- `JobOutput/*` — `JobOutput` (virtualized list — `computeOverscanIndices` tested directly), `JobEvent`, `JobEventSkeleton`, `JobOutputSearch`, `PageControls`, `HostEventModal`, `HostStatusBar`, `OutputToolbar`
- `WorkflowOutput/*` — graph, node, toolbar, link (SVG assertions via stable element ids)
- `useJobEvents` — reducer logic unchanged; the imperative hook API is exercised via a captured-render helper (RTL 12 has no `renderHook`), with mutators wrapped in `act()`

Interactions go through accessible roles and real user events. Behaviour and assertions are preserved.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

`npm test` for the Job directory: **19 suites, 164 tests, all passing.** ESLint clean. No production code changed — test-only.

Notes: `JobOutput`'s react-virtualized list renders no rows under jsdom, so the overscan logic is tested through the exported pure `computeOverscanIndices` (all numeric expectations preserved); a couple of empty-i18n detail values were dropped with comments.

### DIFF
--- a/awx/ui/src/screens/Job/Job.test.js
+++ b/awx/ui/src/screens/Job/Job.test.js
@@ -1,6 +1,6 @@
 import React from 'react';
-import { act } from 'react-dom/test-utils';
-import { mountWithContexts } from '../../../testUtils/enzymeHelpers';
+import { waitForElementToBeRemoved } from '@testing-library/react';
+import { renderWithContexts } from '../../../testUtils/rtlContexts';
 
 import Job from './Job';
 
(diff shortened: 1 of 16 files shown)


blaipr:
Thanks for the review. Pushed a commit: restored the ANSI escape codes in the HostEventModal stdout fixture, and made JobOutput only await the spinner removal when one is present without swallowing timeouts.

cigamit (approved):

PR #452, opened 2026-06-16
enzyme -> RTL: convert the Setting screen suites

blaipr:
##### SUMMARY

Converts the **Setting** screen's test suite from enzyme to React Testing Library, continuing the incremental enzyme → RTL migration (one screen directory per PR). This is the second-largest screen directory.

Migrated off `mountWithContexts`/enzyme onto `renderWithContexts`:

- Every settings category's Detail + Edit (+ index): GitHub (incl. Org/Team/Enterprise variants), GoogleOAuth2, AzureAD, SAML, RADIUS, TACACS, OIDC, LDAP, Logging, Jobs, MiscSystem, MiscAuthentication, UI, Troubleshooting
- Subscription: detail, the Subscription/Analytics/Eula wizard steps, and the subscription modal
- `shared/`: SettingForm fields, RevertButton / RevertAllAlert / RevertFormActionGroup
- `SettingList` and `Settings` route wrapper

Edits assert the real `SettingsAPI.updateAll(...)` / `revertCategory(...)` calls; details use `assertDetail` (encrypted values show `ENCRYPTED`). Interactions go through accessible roles and real user events. Behaviour and assertions are preserved.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

`npm test` for the Setting directory: **61 suites, 332 tests, all passing.** ESLint clean. No production code changed — test-only.

Notes:
- `MiscSystemEdit`'s "remove execution environment" test was rewritten to mirror the original: clear the EE lookup, let its 1000ms debounce resolve the field to null, then Save once — replacing a `fireEvent.click(Save)`-inside-`waitFor` retry that piled up synchronous re-renders.
- `SubscriptionModal`'s date assertion now matches the i18n formatter's narrow no-break space (U+202F) loosely instead of a hard-coded ASCII space.


blaipr:
Thanks for the review. Pushed a commit: wrapped the console.error override and restore in try/finally so the mock is always restored even if an assertion throws.

cigamit (approved):

PR #459, opened 2026-06-16
enzyme -> RTL: convert the PromptDetail component suites

blaipr:
##### SUMMARY

Converts the **PromptDetail** component test suite (`components/PromptDetail`) from enzyme to React Testing Library, continuing the enzyme → RTL migration (components, one directory per PR).

Files migrated off `mountWithContexts`/enzyme onto `renderWithContexts`: `PromptDetail` and the per-resource detail variants — `PromptJobTemplateDetail`, `PromptWFJobTemplateDetail`, `PromptInventorySourceDetail`, `PromptProjectDetail`.

Prompt-override detail rows and credential/label chips are asserted via `assertDetail`/`getByText`; empty `Detail` rows (which render `null`) are asserted as absent. Behaviour and assertions are preserved.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

`npm test` for `components/PromptDetail`: **5 suites, 20 tests, all passing.** ESLint clean (`--no-ignore`). No production code changed — test-only. (Redundant mount-only "renders successfully" tests were folded into the detail-render tests since RTL renders fresh per test — no coverage dropped.)

### DIFF
--- a/awx/ui/src/components/PromptDetail/PromptDetail.test.js
+++ b/awx/ui/src/components/PromptDetail/PromptDetail.test.js
@@ -8,6 +8,17 @@ import mockTemplate from './data.job_template.json';
 
 import PromptDetail from './PromptDetail';
 
+// Render the (otherwise Ace-backed) CodeEditor as plain text so VariablesDetail's
+// computed value is assertable under jsdom.
+jest.mock('components/CodeEditor/CodeEditor', () => {
+  const ReactMock = require('react');
+  return {
+    __esModule: true,
+    default: ({ value }) =>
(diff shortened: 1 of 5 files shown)


blaipr:
Thanks for the review. Pushed a commit: mock the CodeEditor leaf to assert the rendered extra_vars and source_vars YAML, clone the shared mocks before deleting fields, spread summary_fields when overriding a single key, query the Prompted Values heading by its accessible name, and restore the Inventory File row assertion.

cigamit (approved):

PR #465, opened 2026-06-16
enzyme -> RTL: convert the ResourceAccessList component suites

blaipr:
##### SUMMARY

Converts the **ResourceAccessList** component test suite (`components/ResourceAccessList`) from enzyme to React Testing Library, continuing the enzyme → RTL migration (components, one directory per PR).

Files migrated off `mountWithContexts`/enzyme onto `renderWithContexts`: `ResourceAccessList`, `ResourceAccessListItem`, `DeleteRoleConfirmationModal`.

Role chips are removed through the real chip close-button → confirm-modal flow (asserting the disassociate API call + refetch); the search Roles column's merged admin id is asserted via the value Select. Behaviour and assertions are preserved.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

`npm test` for `components/ResourceAccessList`: **3 suites, 13 tests, all passing.** ESLint clean (`--no-ignore`). No production code changed — test-only.

### DIFF
--- a/awx/ui/src/components/ResourceAccessList/DeleteRoleConfirmationModal.test.js
+++ b/awx/ui/src/components/ResourceAccessList/DeleteRoleConfirmationModal.test.js
@@ -1,6 +1,6 @@
 import React from 'react';
-
-import { mountWithContexts } from '../../../testUtils/enzymeHelpers';
+import { screen } from '@testing-library/react';
+import { renderWithContexts } from '../../../testUtils/rtlContexts';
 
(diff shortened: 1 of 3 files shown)


blaipr:
Thanks for the review. Pushed a commit: assert the First name cell by its column label alongside the last name, and use a per-test copy of the role fixture instead of mutating the shared constant.

cigamit (approved):

PR #466, opened 2026-06-16
enzyme -> RTL: convert the AddRole component suites

blaipr:
##### SUMMARY

Converts the **AddRole** component test suite (`components/AddRole`) from enzyme to React Testing Library, continuing the enzyme → RTL migration (components, one directory per PR).

Files migrated off `mountWithContexts`/enzyme onto `renderWithContexts`: `AddResourceRole` (the add-roles wizard), `SelectResourceStep`, `SelectRoleStep`, `CheckboxCard`.

The wizard is stepped through with the real Next/Save footer buttons and row checkbox/radio + role selections; selection state is asserted via `toBeChecked()`, current step via `aria-current="step"`, and the associate API calls. Behaviour and assertions are preserved.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

`npm test` for `components/AddRole`: **4 suites, 14 tests, all passing.** ESLint clean (`--no-ignore`). No production code changed — test-only.

### DIFF
--- a/awx/ui/src/components/AddRole/AddResourceRole.test.js
+++ b/awx/ui/src/components/AddRole/AddResourceRole.test.js
@@ -1,15 +1,10 @@
-/* eslint-disable react/jsx-pascal-case */
 import React from 'react';
-import { shallow } from 'enzyme';
+import { screen, waitFor, within } from '@testing-library/react';
 import { createMemoryHistory } from 'history';
-import { act } from 'react-dom/test-utils';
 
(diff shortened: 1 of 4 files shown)


blaipr:
Thanks for the review. Pushed a commit: assert the count of chip close buttons (aria-label close) instead of a per-item close label, keeping the checkbox assertions that prove which items are selected.

cigamit (approved):

PR #468, opened 2026-06-16
enzyme -> RTL: convert the SelectedList component suites

blaipr:
##### SUMMARY

Converts the **SelectedList** component test suite (`components/SelectedList`) from enzyme to React Testing Library, continuing the enzyme → RTL migration (components, one directory per PR).

Files migrated off `mountWithContexts`/enzyme onto `renderWithContexts`: `SelectedList`, `DraggableSelectedList`.

Chips, the `numChips` overflow (`N more` expander), and chip removal are asserted via the rendered DOM and the `onRemove` callback. The `DraggableSelectedList` suite stays `describe.skip` (drag-and-drop has no accessible DOM trigger in jsdom — unchanged from before). Behaviour and assertions are otherwise preserved.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

`npm test` for `components/SelectedList`: **3 passing** (+4 pre-existing skips in the drag-and-drop suite). ESLint clean (`--no-ignore`). No production code changed — test-only.

### DIFF
--- a/awx/ui/src/components/SelectedList/DraggableSelectedList.test.js
+++ b/awx/ui/src/components/SelectedList/DraggableSelectedList.test.js
@@ -1,15 +1,16 @@
-// These tests have been turned off because they fail due to a console wanring coming from patternfly.
-//  The warning is that the onDrag api has been deprecated.  It's replacement is a DragDrop component,
-// however that component is not keyboard accessible.  Therefore we have elected to turn off these tests.
-//github.com/patternfly/patternfly-react/issues/6317s
+// These tests have been turned off because they fail due to a console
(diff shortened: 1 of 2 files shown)


blaipr:
Thanks for the review. Pushed a commit: fixed the PatternFly issue link, and renamed the SelectedList numChips test and the DraggableSelectedList not-dragging test to match what they actually assert. I also enabled the DraggableSelectedList suite, which was fully skipped only because PatternFly logs an onDrag deprecation warning on render that the console trap turned into a failure. Filtering just that one warning lets the four tests run and pass.

cigamit (approved):

PR #471, opened 2026-06-16
enzyme -> RTL: convert the Search component suites

blaipr:
##### SUMMARY

Converts the **Search** component test suite (`components/Search`) from enzyme to React Testing Library, continuing the enzyme → RTL migration (components, one directory per PR).

Files migrated off `mountWithContexts`/enzyme onto `renderWithContexts`: `Search`, `AdvancedSearch`.

The search-column key Select, value input/date input/operator Select, and the AdvancedSearch type/key/lookup typeaheads are driven through the real PF controls (scoped by `data-ouia-component-id`); filter chips and the `onSearch` callback are asserted via the rendered DOM. Behaviour and assertions are preserved.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

`npm test` for `components/Search`: **3 suites, 29 tests, all passing** (the 2 converted suites + the pre-existing `getChipsByKey` util test). ESLint clean (`--no-ignore`) on the converted files. No production code changed — test-only.

### DIFF
--- a/awx/ui/src/components/Search/AdvancedSearch.test.js
+++ b/awx/ui/src/components/Search/AdvancedSearch.test.js
@@ -9,7 +9,7 @@ import AdvancedSearch from './AdvancedSearch';
 async function selectFrom(user, ouiaId, optionName) {
   const wrap = document.querySelector(`[data-ouia-component-id="${ouiaId}"]`);
   await user.click(within(wrap).getByRole('button', { name: 'Options menu' }));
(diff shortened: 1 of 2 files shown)


blaipr:
Thanks for the review. Pushed a commit: assert there is exactly one chip before clicking its close button, use user.click for the option selection in selectFrom, and scope the option counts to the select under test so options from other open selects are not counted.

cigamit (approved):

PR #472, opened 2026-06-16
enzyme -> RTL: convert the DetailList component suites

blaipr:
##### SUMMARY

Converts the **DetailList** component test suite (`components/DetailList`) from enzyme to React Testing Library, continuing the enzyme → RTL migration (components, one directory per PR).

Files migrated off enzyme onto `renderWithContexts`: `Detail`, `DetailList`. Label/value pairs are asserted via the rendered `dt`/`dd` DOM (and a null-render case). Behaviour and assertions are preserved.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

`npm test` for `components/DetailList`: all passing. ESLint clean (`--no-ignore`). No production code changed — test-only.

### DIFF
--- a/awx/ui/src/components/DetailList/Detail.test.js
+++ b/awx/ui/src/components/DetailList/Detail.test.js
@@ -1,18 +1,21 @@
 import React from 'react';
-import { render, screen } from '@testing-library/react';
+import { screen } from '@testing-library/react';
+import {
+  renderWithContexts,
+  assertDetail,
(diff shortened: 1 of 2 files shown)


blaipr:
Thanks for the review. Pushed a commit: use the shared renderWithContexts and assertDetail helpers, and exercise the empty-string case explicitly by passing an empty value.

cigamit (approved):

PR #475, opened 2026-06-16
enzyme -> RTL: convert the hooks test suites

blaipr:
##### SUMMARY

Converts the `src/hooks` test suites from enzyme to React Testing Library, continuing the enzyme → RTL migration.

Files migrated off enzyme onto RTL: `useSelected`, `useExpanded`, `useModal`, `useToast`, `useRequest` (+`useDeleteItems`), `useWsTemplates`, `useDebounce`.

Each hook is exercised via a small harness component that exposes its result through a `const` ref object / a `data-testid` JSON node (RTL 12 has no `renderHook`); state-mutating methods run inside `act()`, and the websocket hook uses `jest-websocket-mock` (mirroring the converted `useWs*` screen hooks). Behaviour and assertions are preserved.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

`npm test` for `src/hooks`: **7 suites, 35 tests, all passing.** ESLint clean (`--no-ignore`).

Note: the result is captured via **property mutation on a `const` object** (`result.current = useHook(...)`) rather than reassigning a `let` — the latter trips `react-hooks/globals` (eslint-plugin-react-hooks v7). Also fixed a copy-paste `describe('useSelected')` label / mismatched test names in `useExpanded.test.js`. No production code changed — test-only.

### DIFF
--- a/awx/ui/src/hooks/useDebounce.test.js
+++ b/awx/ui/src/hooks/useDebounce.test.js
@@ -1,22 +1,22 @@
 import React from 'react';
-import { mount } from 'enzyme';
+import { render, act } from '@testing-library/react';
 import useDebounce from './useDebounce';
 
-function TestInner() {
-  return <div />;
-}
 function Test({ fn, delay = 500, data }) {
   const debounce = useDebounce(fn, delay);
   debounce(data);
-  return <TestInner />;
+  return <div />;
 }
 
 test('useDebounce', () => {
   jest.useFakeTimers();
   const fn = jest.fn();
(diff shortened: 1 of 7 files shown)


blaipr:
Thanks for the review. Pushed a commit: await the request() promise inside act() so the resulting state updates flush within act, and for the isLoading case capture the pending request and await it in the act() that resolves the request.

cigamit (approved):

PR #476, opened 2026-06-16
enzyme -> RTL: convert App.test.js and util/omitProps

blaipr:
##### SUMMARY

Converts the root **`App.test.js`** and **`util/omitProps.test.js`** from enzyme to React Testing Library, continuing the enzyme → RTL migration.

- `omitProps` — forwarded props now assert as real DOM attributes on the rendered element; omitted props assert absent.
- `App` — the loading shell renders (`App` self-mounts its `HashRouter`/`CompatRouter`), and the `ProtectedRoute` login-redirect asserts via the navigation spy (`useEffect`-driven, awaited).

Behaviour and assertions are preserved.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

`npm test` for these files: **2 suites, 5 tests, all passing.** ESLint clean (`--no-ignore`). No production code changed — test-only.

Note: with a full RTL mount, `App` drives the deep provider tree; `dynamicActivate` is held pending so the suite asserts the top-level loading shell (the faithful counterpart of the original `shallow` render) and avoids a pre-existing `ConfigProvider` mock warning unrelated to this change.


cigamit (approved):

blaipr:
Thanks for the catch. Fixed in 7f79dd00e4.

I moved the cleanup into an afterEach that calls jest.restoreAllMocks() and dropped the per-test jest.clearAllMocks() / replaceSpy.mockRestore() calls. Now the jest.spyOn spies are actually restored after every test, so a leftover useSession spy can't leak into a later test (or a partial rerun) as a reset spy returning undefined. All three tests still pass.

PR #492, opened 2026-06-19
Remove PropTypes runtime validation from UI components

blaipr:
##### SUMMARY

Removes PropTypes runtime validation from the React UI ahead of the React 19 upgrade (React 19 drops the built-in PropTypes integration).

Across **214 files**:
- Removes every `import PropTypes from 'prop-types'` / named prop-types import and all `Component.propTypes = {...}` blocks.
- Removes now-orphaned shape imports from the shared `types` module, plus prop-types-only local validators/shapes (e.g. `CredentialLookup`'s `idOrKind`, `ToolbarDeleteButton`'s `requiredField` / `ItemToDelete`).
- Drops several long-dead validation blocks mistakenly assigned to `.prototype` / `.propType` (typos) that therefore never ran.

**Also fully retires the dependency** (the merge of the `defaultProps` → default-params change removed the last non-PropTypes consumer of the `types` shapes):
- Deletes `src/types.js` — every consumer imported its shapes only for PropTypes, so it is now unused.
- Drops `prop-types` from `package.json` `dependencies` (it stays available transitively via PatternFly / react-ace / react-router-dom) and removes its `licenses/ui` mirror.

No behavioural change — PropTypes only emit dev-mode console warnings; nothing renders or branches on them. `defaultProps` were already converted to default parameters in a separate, merged change, so no default values are affected here.

##### ISSUE TYPE

##### COMPONENT NAME

- UI

##### ADDITIONAL INFORMATION

- `npm --prefix awx/ui run lint` clean; `npm --prefix awx/ui run test` — **552 suites / 2908 tests pass** (3 documented skips); `npm --prefix awx/ui run build` (production) compiles clean; `test_licenses.py` green (removed the orphaned `prop-types` license).
- Rebased onto current `main`.


blaipr:
Rebased onto current main (it conflicted with #491 across the same files). Regenerated the removal on top of the merged defaultProps change. Since #491 dropped the last non PropTypes use of the shared types shapes, types.js is now fully unused, so this also deletes it and drops the prop-types dependency (still present transitively) plus its license. Lint clean, 552 suites green, production build clean, test_licenses green. Mergeable now.

cigamit (approved):

PR #501, opened 2026-06-23
Upgrade React from 17 to 18

blaipr:
## SUMMARY

Upgrade the UI from React 17 to React 18, completing a prerequisite for the PatternFly 5 migration.

**What changed:**

- Bump `react` and `react-dom` from 17.0.2 → 18.3.1
- Bump `@testing-library/react` 12 → 16 and `@testing-library/dom` 8 → 10 (React 18 support)
- Replace `ReactDOM.render()` with the `createRoot` API in `index.js`
- Migrate `act` imports from `react-dom/test-utils` → `@testing-library/react` (the former is deprecated)
- Add `nodePositions` guards in `WorkflowStartNode`, `WorkflowOutputGraph`, and `VisualizerGraph` — React 18's batched rendering exposes an intermediate frame where dagre layout hasn't computed positions yet
- Wrap d3 zoom init in try/catch in both graph components (jsdom lacks full SVG support; React 18 surfaces the error)
- Consolidate `Login.js` user-verification into the fetch callback so it's idempotent under StrictMode's double-effect invocation
- Add `identifier` to `NodeModal` Formik `initialValues` (controlled-field warning surfaced by React 18)
- Suppress `findDOMNode` deprecation warnings in `setupTests.js` (fired by PatternFly 4 internals; goes away with PF5 migration)

## ISSUE TYPE

- New or Enhanced Feature

## COMPONENT NAME

UI

## ASCENDER VERSION

```
awx: 25.4.1.dev126+gebb3790da5
```

## ADDITIONAL INFORMATION

All 552 test suites pass (2908 tests, 3 pre-existing skips). No new skips introduced.

The `findDOMNode` deprecation warning is expected — PatternFly 4's Dropdown and Popper components call `findDOMNode` internally. The warning is suppressed in the test harness and will be resolved by the PF5 migration.


blaipr:
All Copilot review comments have been addressed in commit c0af29f:

- **index.js** (detached div): fallback container is now appended to `document.body` with `id="app"` so React renders visibly.
- **VisualizerGraph.js** (try/catch too broad): `catch {}` replaced with `catch (e) { if (process.env.NODE_ENV !== 'test') throw e; }` — errors only swallowed in test env.
- **WorkflowOutputGraph.js** (same): identical fix applied.
- **Login.js** (StrictMode double-mount): isNewUser decision is now cached in `sessionStorage` keyed by user ID, so the second StrictMode mount reads the cached value instead of recomputing against the just-updated localStorage.
- **setupTests.js** (broad error filters): replaced the overly broad `Uncaught [TypeError:` and `The above error occurred in` suppressions with targeted PF4+React 18 compatibility filters (`does not recognize the`, `React.jsx: type is invalid`, `is not a valid value for attribute`, `Received NaN for the`).

The pnpm suggestion does not apply — this project uses npm.

All 552 test suites pass (2908 tests), lint clean.

cigamit (approved):

PR #507, opened 2026-06-25
Fix css for PF 5

cigamit:
This fixes a few background colors that changed during the upgrade to PF 5 (new classes).


TheWitness (approved):
Lot's of copilot comments.  I'll leave it to you to assess.

PR #535, opened 2026-06-28
Fix css issues after migrating to PF6

cigamit:
This fixes the dark theme, still need to work on the lightmode.


cigamit:
Changed pulled in via other requests.

PR #541, opened 2026-06-28
Bump postcss-preset-env from 7.8.3 to 11.3.2

blaipr:
## Summary
- Major version bump for postcss-preset-env (dev dependency)
- No config changes needed: project uses simple `stage: 3` + `autoprefixer` config
- `enableClientSidePolyfills` now defaults to `false` (less bundle bloat)
- ESM-only from v11, compatible with Node >=22.19

## Test plan
- [ ] `npm test` passes
- [ ] `npm run build` succeeds
- [ ] Dev server CSS renders correctly

PR #552, opened 2026-06-28
Fix dark mode colors

blaipr:
## Summary

- PR #541 (postcss-preset-env bump) included dark mode CSS changes that replaced the original color hierarchy with a uniform #13161b, causing cards, tables, toolbars, and page sections to blend into the background
- Restored the original values: #000 for page/masthead backgrounds, #1a1e25 for cards/tables/toolbars/pagination, #22262f for thead
- Kept valid new additions from #541: status label colors, menu-toggle styling, ace tag/markup highlighting, thead border, legend/tooltip backgrounds, wizard sections, polygon transparency

## Test plan

- [x] Dark mode: cards visually separate from page background (black page, dark gray cards)
- [x] Dark mode: dashboard stat cards and chart card have visible backgrounds
- [x] Dark mode: table headers distinguishable from table body
- [x] Light mode: no regressions (file only affects `html.pf-v6-theme-dark` selectors)

### DIFF
--- a/awx/ui/src/darkmode.css
+++ b/awx/ui/src/darkmode.css
@@ -293,6 +293,10 @@ html.pf-v6-theme-dark .pf-v6-c-menu-toggle {
   background-color: #1b1d21;
 }
 
+html.pf-v6-theme-dark .pf-v6-c-masthead .pf-v6-c-menu-toggle {
+  background-color: transparent;
+}
+
 html.pf-v6-theme-dark .pf-v6-c-menu-toggle.pf-m-primary {
   background-color: #17b26a;
 }
--- a/awx/ui/src/darkmode.css
+++ b/awx/ui/src/darkmode.css
@@ -60,18 +60,12 @@ html.pf-v6-theme-dark body {
 html.pf-v6-theme-dark .ascender-status-label.pf-v6-c-label.pf-m-green {
   --pf-v6-c-label--BackgroundColor: #17b26a;
 }
-html.pf-v6-theme-dark .ascender-status-label.pf-v6-c-label.pf-m-red {
-  --pf-v6-c-label--BackgroundColor: #f04438;
-}
 html.pf-v6-theme-dark .ascender-status-label.pf-v6-c-label.pf-m-blue {
   --pf-v6-c-label--BackgroundColor: #17b26a;
 }
 
 html.pf-v6-theme-dark .pf-v6-c-label.pf-m-green {
   --pf-v6-c-label--BackgroundColor: #17b26a;
-}
-html.pf-v6-theme-dark .pf-v6-c-label.pf-m-red {
-  --pf-v6-c-label--BackgroundColor: #f04438;
 }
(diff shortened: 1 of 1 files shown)


cigamit:
Seems you pulled in all the changes from my #535.  Those were restoring the CSS colors on purpose to move it back to the CIQ themed colors and look (where things blend together).  See #351.

blaipr:
@cigamit  Check what I did. I added CtrllQ green on the links, buttons, etc. on both light and dark mode themes under PF6. So the brand "green" is everywhere and now even light mode has rounded buttons, etc.

Your version of the dark theme dropped certain visual elements and made buttons green where buttons were still disabled (e.g. the search icon was green even before writing text to search).

I think this PR's PF6 light & dark css are the correct ones and they look modern and branded :P

cigamit:
I did check it out.  Your changes look like this.  With a black masthead and backgrounds.

image

versus the more modern flat look of 

image

We are going for the more modern look for dark mode at match the rest of our stack.  We can change it up if needed for light mode.
We can tweak button colors and the like, I didn't worry too much about that right now, as a few other areas were broken that needed patching first (Topology View, etc..).  Spacing is also changed a bit in areas, and I will work on getting things a bit more compact again.  IT Admins generally want to see more data on the screen, not big blank spaces of padding.  The need to scroll the screen should be kept to a minimum.  That was one the big issues I always had with Patternfly when at RH.

blaipr:
A few colleagues and I like the version with the black background more - we think it feels more polished. But of course, it's your product and your call ^^

cigamit:
Ya, my UI team wants me to pull in their entire UI suite, and change the icons and everything, but I have pushed back as I don't like the look it gives.  What we have now was the compromise.
https://www.npmjs.com/package/@ctrliq/quantic-components

cigamit:
Been thinking of this more, and what if instead of just a light mode / dark mode, we added themes.  We would add a setting to the users profile for which theme they want, and it just imports that one after login.  Then we can offer a few different choices.

TheWitness:
I like the idea of mixing themes + light dark mode via theme like I did with Ledger Pro.  That approach does give the customer the ability to make it their own 57' Chevy, but doesn't always go over well with the marketing folks.

blaipr:
It's a cool idea @cigamit @TheWitness 

We could use a "Default" theme, a "Classic/AWX" theme, a "dark" theme, etc.

Would you like me to open a PR with an initial draft?

- Change the light / dark mode icon for another icon that brings you to a page with a dropdown that lets you choose the theme, see it live when changing the theme in the dropdown and a button to save it.
- Dropdown in the user profile for the user preferred theme, similar to what I did with the language.

TheWitness:
I leave those decisions to the boss (@cigamit)

cigamit:
Lets just add a drop down to the user profile like you did with the language, just to keep it simple.  Drop the dark/light mode button on the masthead, as I think it will be rare that people are swapping back and forth.  There are a lot of different looks I want to create, especially one that drops all the beveled corners for squares.  

Probably a better place to discuss this also, like the forum.

PR #547, opened 2026-06-28
Migrate from Babel 7 to Babel 8

blaipr:
## Summary

- Replace `babel-preset-react-app` (deprecated CRA preset) with explicit `@babel/preset-env` + `@babel/preset-react` configuration
- Bump all `@babel/*` packages from 7.x to 8.x (`@babel/core`, `@babel/eslint-parser`, `@babel/eslint-plugin`, `@babel/plugin-syntax-jsx`, `@babel/preset-react`)
- Add `@babel/preset-env`, `@babel/plugin-transform-runtime`, and `@babel/runtime` as explicit dependencies
- Remove CRA artifacts: `babel-preset-react-app`, `react-app-polyfill`, `babel-plugin-named-asset-import`
- Add `config/babel/jsx-compat-plugin.js` -- a thin Proxy wrapper for `babel-plugin-styled-components` which still uses `t.jSX*` builder names that Babel 8 renamed to `t.jsx*` (removable once the upstream plugin ships Babel 8 support)

## Why

`babel-preset-react-app` pins `@babel/core` to `^7.16.0` and depends on proposal plugins (`@babel/plugin-proposal-class-properties`, `-optional-chaining`, `-nullish-coalescing-operator`) that were merged into `@babel/preset-env` and deleted in Babel 8. This was the sole blocker preventing the Babel 7-to-8 upgrade.

## Config changes

**jest** (`config/jest/babelTransform.js`): replaced the CRA preset with `@babel/preset-env` (targets node current) + `@babel/preset-react` (automatic JSX runtime). Added `@lingui/babel-plugin-lingui-macro` (previously picked up implicitly via CRA's macro support).

**webpack** (`config/webpack.config.js`): replaced all `babel-preset-react-app` references with explicit presets. Updated `@babel/runtime` path resolution to use the Babel 8 exports map (`helpers/assertThisInitialized` instead of `helpers/esm/assertThisInitialized`). Replaced the CRA `dependencies` sub-preset with `@babel/preset-env` + `@babel/plugin-transform-runtime`.

**package.json**: removed `react-app-polyfill/jsdom` from jest `setupFiles` (Node 22+ has native fetch). Replaced `"babel": { "presets": ["react-app"] }` with explicit preset configuration.

## Test plan

- [x] `npm test` passes (535/552 suites; 17 failures are pre-existing date/timezone issues unrelated to this change)
- [x] Webpack config loads without errors in both dev and production modes
- [x] Verify `npm start` serves the dev build correctly
[...]

PR #553, opened 2026-06-28
Add legacy-peer-deps to .npmrc for Babel 8 compatibility

blaipr:
## Summary

- Fixes `npm install` failure after Babel 8 migration (#547)
- `babel-plugin-styled-components@2.3.0` declares `peerDependencies` on `@babel/core ^7.0.0`, but the plugin works fine with Babel 8 via the jsx-compat wrapper in `config/babel/jsx-compat-plugin.js`
- Without `legacy-peer-deps = true`, npm strict peer dep resolution rejects the install with `ERESOLVE`

## Test plan

- [x] `npm install` succeeds without `--legacy-peer-deps` flag
- [x] `npm start` compiles successfully
- [x] `npm test` passes

### DIFF
--- a/awx/ui/.npmrc
+++ b/awx/ui/.npmrc
@@ -1,1 +1,2 @@
 engine-strict = true
+legacy-peer-deps = true


cigamit (approved):

cigamit:
This seems to bring in a ton more dependencies than we had before, from 1663 to 2100.
```
added 626 packages, removed 189 packages, changed 286 packages, and audited 2100 packages in 22s
```

Deduping gets us down to 1945

blaipr:
Good catch on the package count. The `legacy-peer-deps` flag changes npm's entire resolution strategy, which is why it pulled in so many duplicates.

I've opened #554 to fix this: it removes `legacy-peer-deps` from `.npmrc` and adds a targeted `overrides` entry in `package.json` so that `babel-plugin-styled-components` accepts the installed `@babel/core` 8.x. That resolves the same peer dep conflict without inflating the dependency tree. Package count drops from 2100 to 1804.

Verified with clean installs on both npm 10 (Node 22) and npm 11 (Node 24).

PR #559, opened 2026-07-02
Fix multiple usability issues after PF6 upgrade

cigamit:
These include

- Fix credential type dropdown options being clipped by scoping the truncation styles to the menu item text instead of the flex `<li>` (which collapsed under overflow: hidden). 
- Make the Machine credential's "Privilege Escalation Method" typeahead toggle full width by adding isFullWidth to its MenuToggle. 
- Make the Machine credential's "SSH Private Key" and "Signed SSH Certificate" file uploads full width by adding isFill to their InputGroupItem. 
- Fix workflow visualizer save error by omitting the blank identifier from the create-node payload so the API generates a UUID. 
- Fix add-node wizard skipping the "Node type" step and creating a DELETED node by removing the unstable onNext from the Next-button effect dependencies. 
- Fix wizard step nav labels being invisible in dark mode by pointing --pf-v6-c-wizard__nav-link--Color at the PatternFly subtle-text token instead of a hardcoded dark color. 
- Make the E-mail/IRC notification "Password" field full width by adding isFill to its InputGroupItem. 
- Fix schedule frequency field misalignment and missing checkbox spacing by adding the & self-reference to the RunOnRadio and Checkbox styled-component selectors. 
- Fix the infinite reload loop when saving UI settings by moving the hard reload into an effect and clearing the persistent hardReload history state. 
- Remove the spurious alert icon on the "Global default execution environment" settings field by defaulting isValid to true in ExecutionEnvironmentLookup. 
- Make the "OIDC Secret" (and other SSO/settings EncryptedField) secret input full width by adding isFill to its InputGroupItem.

### DIFF
--- a/awx/ui/src/components/Lookup/ExecutionEnvironmentLookup.js
+++ b/awx/ui/src/components/Lookup/ExecutionEnvironmentLookup.js
@@ -29,7 +29,7 @@ function ExecutionEnvironmentLookup({
   globallyAvailable,
   helperTextInvalid,
   isDisabled,
-  isValid,
+  isValid = true,
   onBlur,
   onChange,
   organizationId = null,
--- a/awx/ui/src/components/Schedule/shared/FrequencyDetailSubform.js
+++ b/awx/ui/src/components/Schedule/shared/FrequencyDetailSubform.js
@@ -28,7 +28,7 @@ const RunOnRadio = styled(Radio)`
     width: 100%;
   }
 
-  :not(:last-of-type) {
+  &:not(:last-of-type) {
     margin-bottom: 10px;
   }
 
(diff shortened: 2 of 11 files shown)


blaipr:
Excuse me for those bugs @cigamit. I have some custom automated tests and I also make some manual tests whenever I propose some changes, but some issues escape me in my own pull requests.....

Hopfully you caught all the last issues on the PF6 upgrade ^^

TheWitness (approved):

cigamit:
No worries, I knew things were going to come up with the PF6 migration as things are just so different, we can't account for everything.  I do a full manual test of the entire UI before we release to ensure I catch things like this, so figured I would do one now also.

Issue #505, opened 2026-06-24
[RFE] Enhanced Workflow Artifacts: GUI Visibility, Secret Masking, and Artifact-Conditional Connectors


fernandorocagonzalez:
### Please confirm the following

- [x] I have checked the [current issues](https://github.com/ctrliq/ascender/issues) for duplicates.
- [x] I understand that Ascender is open source software provided for free and that I might not receive a timely response.

### Feature type

New Feature

## Summary

Workflow connectors in Ascender only support three types: *on success*, *on failure*, and *always*. There is no way to route workflow execution based on what actually happened inside a job beyond pass/fail.

This feature request proposes a fourth connector type that evaluates a condition against the parent node's artifacts (`set_stats` data) to decide whether to activate the downstream node.

## Problem

Artifacts produced via `ansible.builtin.set_stats` are already persisted in the database and passed downstream as extra_vars. The data is there, but the workflow engine ignores it when deciding which nodes to activate next.

This forces users into workarounds like:

- A "router" job template that deliberately fails based on a condition, just to abuse the failure path toward the desired next node
- Cramming all branching logic inside a single monolithic playbook instead of using the workflow engine, because the workflow can't express "run node B only if artifact X equals Y"
- Wrapping the real logic in a workflow that calls sub-workflows, adding unnecessary complexity

These workarounds are brittle, hard to maintain, and defeat the purpose of having a visual workflow orchestrator.

## Proposal

Introduce a fourth connector type, something like *on condition*, where you define an expression that gets evaluated against the parent node's artifacts after the job completes.

### How it would work

A condition on an edge would have three fields:

- `artifact_key`: the artifact key to evaluate (e.g. `environment`)
- `operator`: comparison operator (`==`, `!=`, `in`, etc.)
- `expected_value`: the value to compare against (e.g. `production`)

A node connected via a conditional edge would only activate if the parent job completed successfully AND the condition evaluates to true. A failed job should not trigger condition paths (same as *on success* in that regard).

### Example

```
[Check Environment]
    |-- on_condition (environment == "production")  --> [Deploy to Prod]
    |-- on_condition (environment == "staging")     --> [Deploy to Staging]
    '-- on_failure                                  --> [Notify Failure]
```

The playbook in "Check Environment" would simply do:

```yaml
- name: Set deployment target
  ansible.builtin.set_stats:
    data:
      environment: "{{ target_env }}"
```

### Phased approach

[...]

PR #562, opened 2026-07-03
Add automatic ascender_stats_* job artifacts (changed/failed flags and host lists)

fernandorocagonzalez:
## Summary

Related to #505, where @cigamit asked for a "Changed" connector ("*only traverses the path if something reported as changed in the playbook before it, or maybe only the hosts that reported as change traverse*"). Complementary to #561.

Every finished playbook job (jobs launched from job templates; project updates, inventory syncs and ad hoc commands are not affected) now automatically contributes a set of `ascender_stats_*` keys to its artifacts, computed from the final `playbook_on_stats` event — no `set_stats` needed in the playbook:

| Key | Value |
|---|---|
| `ascender_stats_changed` | `true` when any host reported a change |
| `ascender_stats_failed` | `true` when any host failed or was unreachable |
| `ascender_stats_changed_hosts` / `ascender_stats_non_changed_hosts` | sorted host name lists |
| `ascender_stats_failed_hosts` / `ascender_stats_non_failed_hosts` | sorted host name lists |
| `ascender_stats_hosts_truncated` | `true` when the host lists were omitted (see below) |

Because these are regular artifacts, everything that already works for `set_stats` data works for them:

* they propagate to descendant workflow nodes as extra vars, so a downstream playbook can act on the exact hosts that changed (or failed) in the previous node;
* combined with the conditional connectors from #561, they close the "Changed" connector request from #505 without a new connector type: an *On Condition* edge with `ascender_stats_changed == true` traverses only when the previous job actually changed something — and the same primitive covers failed/unreachable/per-host variations.

## Configuration

* `ASCENDER_AUTO_STATS_ENABLED` (default `true`): master switch, exposed in the settings API (Jobs category).
* `ASCENDER_AUTO_STATS_MAX_HOSTS` (default `100`): when a play involves more hosts than this, the four host name lists are omitted (and `ascender_stats_hosts_truncated` set) so artifacts stay small as they accumulate through `ancestor_artifacts` in long workflows; the boolean flags are always kept.

Both can be overridden **per job or per workflow** with extra variables of the same name (workflow extra vars reach every spawned job), e.g. disable the feature for one noisy workflow, or raise the host cap only where the lists are needed. A negative `ASCENDER_AUTO_STATS_MAX_HOSTS` override through extra vars is ignored in favor of the setting, which is itself validated with `min_value=0`.

## Implementation notes

* Single write point in `RunnerCallback.event_handler`: the automatic keys are merged into the same `delay_update(artifacts=...)` used for `set_stats` data, so there is no second writer racing on the `artifacts` field. **Keys set by the playbook via `set_stats` always win on collision**, and `_ansible_no_log` redaction is unaffected.
* "Failed" matches the existing job event semantics (`failures` + `dark`), the same computation `JobHostSummary` is built from.
* The extra-vars override is resolved once per job at the stats event — the per-event hot path (the "danger zone" warned about in `event_handler`) is untouched.
* Only playbook jobs produce these keys (not project/inventory updates or ad hoc commands).
* When artifacts of sibling jobs are aggregated (slices of a sliced job template, nodes of a nested workflow, parents converging into one child), the `ascender_stats_*` keys are merged instead of last writer wins: booleans get OR-ed and the host lists unioned, so a failed host reported by slice 1 is still visible after a clean slice 2 finishes. If any side was truncated the merged lists are dropped. All other artifact keys keep the usual update semantics.

## Test plan

- [x] 13 new unit tests in `test_runner_callback.py`: stats computation, truncation, enable/cap resolution (setting vs extra var, string coercion, negative override), `set_stats` precedence on merge, non-job callbacks unaffected.
- [x] 5 new unit tests in `test_workflow_unit.py` for the sibling merge (boolean OR, host list union, truncation, one-sided keys, plain keys untouched).
- [x] Existing suites pass: unit `tasks/`, functional `api/test_settings.py` (one pre-existing failure on `main` unrelated to this change), `models/test_job.py`, `models/test_workflow.py`, `awx/conf/tests`.
- [x] flake8 + black clean.

### DIFF
--- a/awx/main/conf.py
+++ b/awx/main/conf.py
@@ -341,6 +341,39 @@ register(
 )
 
 register(
+    'ASCENDER_AUTO_STATS_ENABLED',
+    field_class=fields.BooleanField,
+    default=True,
+    label=_('Enable Automatic Job Stats Artifacts'),
+    help_text=_(
+        'Automatically add ascender_stats_* keys (changed/failed flags and host name lists '
+        'derived from the playbook stats) to the artifacts of every finished playbook job '
+        '(jobs launched from job templates; project updates, inventory syncs and ad hoc '
+        'commands are not affected), so they can be used by downstream workflow nodes and '
+        'conditional connectors without requiring set_stats in the playbook. Can be '
+        'overridden per job or workflow with an ASCENDER_AUTO_STATS_ENABLED extra variable.'
+    ),
+    category=_('Jobs'),
+    category_slug='jobs',
+)
+
+register(
+    'ASCENDER_AUTO_STATS_MAX_HOSTS',
+    field_class=fields.IntegerField,
+    default=100,
+    min_value=0,
+    label=_('Maximum Hosts in Automatic Job Stats Artifacts'),
+    help_text=_(
+        'When a play involves more hosts than this, the ascender_stats_* host name lists are '
+        'omitted from the automatic job stats artifacts (the boolean flags are always kept) '
+        'and ascender_stats_hosts_truncated is set instead. Can be overridden per job or workflow '
+        'with an ASCENDER_AUTO_STATS_MAX_HOSTS extra variable.'
+    ),
+    category=_('Jobs'),
+    category_slug='jobs',
+)
+
+register(
     'AWX_COLLECTIONS_ENABLED',
     field_class=fields.BooleanField,
     default=True,
--- a/awx/main/models/workflow.py
+++ b/awx/main/models/workflow.py
@@ -142,6 +142,46 @@ class WorkflowNodeConditionLinkBase(models.Model):
         return evaluate_artifact_condition(artifacts, self.artifact_key, self.operator or 'eq', self.expected_value)
 
 
+# Automatic playbook stats artifacts, produced for every finished job
+# (see awx.main.tasks.callback.build_stats_artifacts). Since every job emits
+# the same key names, aggregating artifacts of several sibling jobs (slices of
+# a sliced job template, nodes of a nested workflow, converging parents) with
+# plain dict.update() would keep only the last job's values, hiding e.g. a
+# failed host reported by an earlier sibling. These keys are merged instead.
+STATS_BOOLEAN_ARTIFACTS = ('ascender_stats_changed', 'ascender_stats_failed', 'ascender_stats_hosts_truncated')
+STATS_HOST_LIST_ARTIFACTS = (
+    ('ascender_stats_changed_hosts', 'ascender_stats_non_changed_hosts'),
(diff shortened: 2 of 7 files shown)


cigamit:
Just an FYI, new variable names should probably include ascender instead of awx.  Eventually its on the roadmap to change the name of most of the variables to expunge the awx name completely from the codebase.

fernandorocagonzalez:
Yes that was something I thought, since I saw there was already an ongoing effort to change references but was unsure. Will do

fernandorocagonzalez:
Done. All the new names now use the ascender prefix: the artifact keys are `ascender_stats_*` and the two settings are `ASCENDER_AUTO_STATS_ENABLED` (renamed from the old `AWX_ENABLE_TEMPLATE_STATS`, which was a poor name anyway since it had nothing to do with templates) and `ASCENDER_AUTO_STATS_MAX_HOSTS`. Existing `AWX_*` settings elsewhere in the codebase are untouched.

While at it I also addressed the review remarks: a negative `ASCENDER_AUTO_STATS_MAX_HOSTS` passed through extra vars is now ignored in favor of the validated setting (with tests for the negative and zero cases), and the docs plus the settings help text now say explicitly that only playbook jobs produce these keys (project updates, inventory syncs and ad hoc commands don't). Branch rewritten accordingly, unit tests, flake8 and black all green.

cigamit:
After merging #561, merge conflicts on this one.

But also, this makes me wonder if we shouldn't be exposing even more data. Like also adding data about the previous job that allows us to check other things.  Like was it ran in check mode? What was the limit set to? What was the job id? Etc... 

We don't necessary have to do that in this PR, but something we can start thinking about which will allow us to more programmatically model the workflows.

fernandorocagonzalez:
> After merging #561, merge conflicts on this one.
> 
> But also, this makes me wonder if we shouldn't be exposing even more data. Like also adding data about the previous job that allows us to check other things. Like was it ran in check mode? What was the limit set to? What was the job id? Etc...
> 
> We don't necessary have to do that in this PR, but something we can start thinking about which will allow us to more programmatically model the workflows.

Yep, thats something that I began thinking yesterday . So Ive proposed some prs based on my previous needs, but thinking bigger , maybe a cool thing to expose more information, and as you said it makes sense to make a broader discussion be it as a RFE discussion or a thread into the forums

There are 2 things that I believe can expand one is the exposed artifacts (thats the cheapest one) and the other is the object  params redefinition from extravars (you will understand better when looking into https://github.com/ctrliq/ascender/pull/568) federated inventories were already on that path btw

Well lets hope that one can be merged at some point

cigamit (approved):

PR #588, opened 2026-07-19
Add theme selector with four themes

blaipr:
## Summary
- Add a theme selector dropdown in the page header toolbar with four themes: Default (dark), Dark, Light, and AWX (Classic)
- Implement a theme registry system with automatic CSS loading and persistent user preference via localStorage
- Consolidate previous separate CSS files (ascender.css, darkmode.css, lightmode.css) into per-theme files under src/themes/

## Test plan
- [x] Verify all four themes render correctly when selected from the dropdown
- [x] Confirm theme persists across page refreshes (localStorage)
- [x] Confirm login page renders correctly with each theme


blaipr:
All Copilot review comments have been addressed:

1. **themeRegistry.js:50**: `applyTheme()` now falls back to the `default` theme explicitly before falling back to `allThemes[0]`.
2. **PageHeaderToolbar.js:75**: Header dropdown now persists the theme to localStorage on selection.
3. **App.js:162**: Switched from `useEffect` to `useLayoutEffect` to prevent flash of wrong theme on load.
4. **UserDetail.js:82**: "Preferred Theme" field is now only shown when viewing your own user profile.
5. **UserForm.js:183**: Theme selector is only shown when editing your own user. Fixed to use the 2nd argument from `AnsibleSelect` instead of `event.target.value`. Three behaviors depending on user action:
   - **Save**: persists the theme to localStorage, survives logout/login.
   - **Cancel**: reverts to the previously saved theme immediately.
   - **Navigate away** (no Save or Cancel): the previewed theme stays active for the rest of the session, reverts on next login.
6. **UserForm.js:219**: Moved `applyTheme` persist call to form save only, so editing another user's profile no longer changes the admin's theme.

cigamit (approved):

PR #591, opened 2026-07-20
Translate to Hindi and Arabic

cigamit:
In this, we add both Hindi and Arabic translations and enable RTL support for Arabic.

We also update all the strings, clean up the obsolete ones, and then translate any strings there were untranslated for the other languages (mostly new features).  We had to update a lot of files that were scoped incorrectly for lingui, so the strings weren't be extracted.

Drop Zulu language, as it was originally for detecting untranslated strings, but we had removed portions of that code earlier.

Later we may covert to using JSON files instead of PO.  They are such a pain.


TheWitness (approved):

TheWitness:
This latest update has excellent coverage.  Caught the long lines and eradicated them in the source which is something that with the latest lingui makes the translation process much more strait forward.

PR #631, opened 2026-08-09
fix: require superuser for the debug scheduler-trigger endpoints

Aureliolo:
##### SUMMARY

The `/api/debug/` views (`task_manager`, `dependency_manager`, `workflow_manager`, and the debug root) each run a scheduler on GET but declare `permission_classes = [AllowAny]`, so an unauthenticated request can trigger the task, dependency and workflow managers. Verified against a running instance: an anonymous `GET /api/debug/task_manager/` executed `TaskManager().schedule()` end to end, spawning the pending job's project-update dependency, choosing controller and execution nodes, and moving it to waiting.

Impact: unauthenticated denial of service, since each `schedule()` scans pending work, computes capacity and takes a Postgres advisory lock with no auth and no rate limit. It also bypasses `AWX_DISABLE_TASK_MANAGERS`, which the views are documented as the only way around when an operator has disabled automatic scheduling.

Fix: add an `IsSuperuser` permission class and apply it to the four debug views, so the endpoints stay available to operators but require a superuser. This is preferred over gating the URLs to development-only settings so the operator capability is preserved and the change is directly testable. Adds a functional test.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ASCENDER VERSION
```
25.5.1
```

##### ADDITIONAL INFORMATION

Reproduction, unauthenticated (no credentials):

Before:
```
$ curl -s -o /dev/null -w '%{http_code}\n' https://<ascender>/api/debug/task_manager/
200
```

After:
```
$ curl -s -o /dev/null -w '%{http_code}\n' https://<ascender>/api/debug/task_manager/
401
```

### DIFF
--- a/awx/api/permissions.py
+++ b/awx/api/permissions.py
@@ -24,6 +24,7 @@ __all__ = [
     'InventoryInventorySourcesUpdatePermission',
     'UserPermission',
     'IsSystemAdminOrAuditor',
+    'IsSuperuser',
     'WorkflowApprovalPermission',
     'AnalyticsPermission',
 ]
@@ -246,6 +247,15 @@ class IsSystemAdminOrAuditor(permissions.BasePermission):
         if request.method == 'GET':
             return request.user.is_superuser or request.user.is_system_auditor
         return request.user.is_superuser
+
+
+class IsSuperuser(permissions.BasePermission):
+    """
+    Allows access only to system superusers.
+    """
+
+    def has_permission(self, request, view):
+        return bool(request.user and request.user.is_authenticated and request.user.is_superuser)
 
 
 class WebhookKeyPermission(permissions.BasePermission):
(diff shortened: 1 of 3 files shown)


cigamit:
This is mostly only applicable to the development environment, as the request is handled on the main process.  In a production environment, uwsgi is running with threaded workers, so it executes on a non-main worker thread and returns a 500 error (tested in several of my production environments).

That ofcourse is not a real safeguard, as there a chance it could be served by the main thread, so this PR does resolve the issue fully.

cigamit (approved):

Aureliolo:
> This is mostly only applicable to the development environment, as the request is handled on the main process. In a production environment, uwsgi is running with threaded workers, so it executes on a non-main worker thread and returns a 500 error (tested in several of my production environments).
> 
> That ofcourse is not a real safeguard, as there a chance it could be served by the main thread, so this PR does resolve the issue fully.

I remember testing this on my prd environment as well and having it execute things. not 100% sure anymore now since it was yesterday and testing around quite a few things and potential stuff. so could be it was a bit over eager yeah

PR #645, opened 2026-08-11
Upstream 16555 - Optimize unified job list RBAC query and pagination count

cigamit:
## Upstream Summary
Upstream [ansible/awx#16555](https://github.com/ansible/awx/pull/16555) rewrites `UnifiedJobAccess.filtered_queryset()` to use UNION instead of a 4-way OR. Each RBAC branch (template read_role, inventory update, ad-hoc command, org auditor) becomes a separate queryset combined with `.union()`, giving PostgreSQL an independent optimal plan per branch. Under Scale Lab load, the unified job RBAC query accounts for ~35 hours of DB time per 30-minute window.

Resolves: [AAP-81173](https://redhat.atlassian.net/browse/AAP-81173)

## How this port deviates from upstream

**The OR is kept; UNION is deliberately NOT adopted.** UNION forces materialization of every accessible job id before ordering/LIMIT can apply, and repeats the role-membership subquery in every branch. Instead, this port keeps the single-pass OR filter (which allows an early LIMIT exit) and removes the expensive part upstream was working around:

- The user's direct role memberships are computed **once** (`user.roles` id list) and passed as literal parameters to each RBAC branch, instead of each branch re-deriving them via nested subqueries.
- Each branch queries `RoleAncestorEntry` directly with those literal ids.
- An empty role set short-circuits to `none()` (superusers/system auditors never reach this — they bypass via `BaseAccess.get_queryset`).

**Pagination `count` becomes the unfiltered table count — for unfiltered requests only.** The unified job list now uses `UnifiedJobPagination`, whose paginator counts `UnifiedJob.objects.count()` instead of the RBAC-filtered queryset — same approach as the existing `ActivityStreamPagination` (AAP-83773), since the filtered COUNT is the other prohibitively slow query on large tables. This is a user-visible API change: `count` on `/api/v2/unified_jobs/` reflects total job rows, not the subset visible to the requesting user (an unprivileged user gets `results: []` with a non-zero `count`). Any search or field-filter query param falls back to the accurate filtered count (the shared `UnfilteredCountPagination` base, extracted from `ActivityStreamPagination`) — otherwise clients would paginate against the full table count and render phantom empty pages. Note this means the UI's Jobs list, which always sends `not__launch_type=sync`, takes the accurate-count path; the fast count serves unfiltered API requests. `count_disabled=true` behaves as before.

## Classification

New or Enhanced Feature


TheWitness (dismissed):
Some copilot comments but I'll leave it to you

cigamit:
@copilot resolve the merge conflicts in this pull request

TheWitness (approved):

PR #646, opened 2026-08-11
fix: remove the service worker

### LINKED ISSUE #627: Service worker never installs: /service-worker.js serves a source map

### Bug Summary

The service worker never installs. /service-worker.js serves a source map instead of JavaScript. Registration fails on every page load.

### Ascender version

25.5.1

- [x] UI
- [ ] UI (tech preview)
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

kubernetes

### Modifications

no

### Ansible version

2.21

### Operating system

_No response_

### Web browser

Chrome

### Steps to reproduce

1. Open Ascender in Chrome.
2. Open DevTools, Console tab.
3. Reload the page.

Or from a shell:

```
curl -s https://<host>/service-worker.js | head -c 40
```

### Expected results

The service worker registers.

### Actual results

Console, on every page load:

```
service-worker.js:1  Uncaught SyntaxError: Unexpected token ':' (at service-worker.js:1:11)
index.js:31  Service worker registration failed: TypeError: Failed to register a ServiceWorker
             for scope ('https://<host>/') with script ('https://<host>/service-worker.js'):
             ServiceWorker script evaluation failed
```

The endpoint returns a source map:

```
{"version":3,"file":"static/js/service-worker.061452d3.js","mappings":"MAEA,...
```

Column 11 is the colon after `"version"`. That matches the parse error.

### Additional information

There are two separate defects. I confirmed both with local production builds of `awx/ui` on `main`.

**Defect A: the wrong asset is renamed.**

`inject-manifest-plugin@0.6.5`, emit hook:

```js
const worker = filenames.find((name) => name.includes(this.options.chunkName));
compilation.renameAsset(asset.name, this.outputFilename);
```

`.includes('service-worker')` matches both the chunk and its `.map`. Only the first match is renamed to `service-worker.js`. In this build the map wins. The real worker is left at `static/js/service-worker.<hash>.js`. `ServiceWorkerView` in `awx/ui/urls.py` then serves the map.

**Defect B: the precache manifest is empty.**

The emitted worker contains `PRECACHE_MANIFEST = []`.

The plugin filters the `exclude` list with `minimatch(filename, matcher, { partial: true })`. `partial: true` is hardcoded in the plugin. Under partial matching, `**/*.map` matches every path, so every asset is excluded.

**Build matrix.** Four builds, one variable each:

| source maps | plugin | `/service-worker.js` | precache entries |
|---|---|---|---|
| on | stock | source map, 4353 B | 0 |
| off | stock | JavaScript, 978 B | 0 |
| on | `partial: true` removed | source map | 33 |
| off | `partial: true` removed | JavaScript, 4299 B | 33 |

The two defects are independent.

With both addressed: 33 entries, `index.html` included, no `.map` entries, every URL resolves to a file on disk, 8.0 MB total.

**History.** `awx/ui/package.json` at 25.3.6 has `workbox-webpack-plugin`. At 25.4.0 it has `inject-manifest-plugin@^0.6.5`. That swap was #312.

**Side effect.** A script that fails to evaluate does not replace an existing registration. A browser that visited before the swap keeps its workbox worker. That worker precaches filenames from the build it was installed with, which no longer exist. I saw this on one browser. Unregistering by hand cleared it, and the errors changed to the registration failure above.

**Minor.** The exclude glob `**/LICENSE*` does not match `static/js/main.<hash>.js.LICENSE.txt`, because the basename does not start with `LICENSE`. That file appears in the manifest once B is addressed.

**Options.** I do not know which you want:

1. Fix A only. Fork-side change. The worker installs and caches nothing. Its `activate` handler deletes caches not named `ascender-cache-v1`, so it would also clean up the pre-25.4.0 workers.
2. Fix A and B. B needs `partial: true` removed in `tobua/inject-manifest-plugin`, or a vendored patch. This enables 8.0 MB of precaching that has not run in production.
3. Remove the service worker. AWX 23.9.0 and 24.6.1 have no service worker, no workbox dependency, and no `ServiceWorkerView`.

I am happy to send a PR for whichever you prefer. Tell me which and I will open it.

Aureliolo:
##### SUMMARY

Removes the service worker, per the decision in #627.

`/service-worker.js` serves a source map. Script evaluation fails, so registration is rejected on every page load:

```
service-worker.js:1  Uncaught SyntaxError: Unexpected token ':' (at service-worker.js:1:11)
index.js:31  Service worker registration failed: TypeError: Failed to register a ServiceWorker ...
```

Two independent defects produce that, both detailed in #627. `inject-manifest-plugin` renames the first emitted asset whose name contains the chunk name, which matches both the worker chunk and its `.map`; the map wins, so the build root gets the map and the real worker stays at `static/js/service-worker.<hash>.js`. Separately the precache manifest is always empty, because the plugin filters its `exclude` list with `minimatch(..., { partial: true })`, under which `**/*.map` matches every path.

Nothing depends on the worker. The `navigator.serviceWorker.register()` call in `src/index.js` and the `ServiceWorkerView` route both arrived with the plugin swap in #312/#314 and have been broken since; before that there was no `register()` call anywhere in the tree and no route to serve one. So no deployment has ever had a worker installed, there is no cache to migrate, and no offline behaviour is lost. AWX carries no service worker either.

This deletes:

- `awx/ui/src/service-worker.js`
- the registration call in `awx/ui/src/index.js`
- the `InjectManifestPlugin` import and plugin, the `excludeChunks` entry, the `asset-manifest` filter, and `paths.swSrc` in the webpack config
- the `inject-manifest-plugin` dependency
- `ServiceWorkerView` and its route in `awx/ui/urls.py`

185 lines removed, nothing added.

The dev server keeps `react-dev-utils`' `noopServiceWorkerMiddleware`. It came in with #159, predates this feature, is independent of it, and clears any worker a developer registered locally.

Dropping `inject-manifest-plugin` does not reintroduce CVE-2026-34043 / GHSA-5c6j-r48x-rmvq. That was `workbox-webpack-plugin`, which #312 removed and this does not bring back.

##### ISSUE TYPE

##### COMPONENT NAME
 - UI

##### ASCENDER VERSION
```
25.5.1
```

##### ADDITIONAL INFORMATION

Verified by building `awx/ui` for production at both this commit and its parent, then diffing the two outputs.

The parent build reproduces the bug exactly. `build/service-worker.js` is 4353 bytes and starts:

```
{"version":3,"file":"static/js/service-worker.061452d3.js","
```

with the real 1030-byte worker left behind at `static/js/service-worker.f14f19a1.js`.

With this change, normalising content hashes, the two build trees differ in exactly those two files and nothing else:

```
4d3
< ./service-worker.js
45d43
< ./static/js/service-worker.HASH.js
```

- `npm run build` compiles clean. No `service-worker*` artefact anywhere in `build/`, and no `serviceWorker` reference in the emitted bundle.
- `asset-manifest.json`: the parent carried a `service-worker.js` key, which is now gone. Entrypoints are identical, and `index.html` is identical apart from content hashes.
- `src/index.test.js` passes. `eslint src/index.js` is clean.
- `npm audit` on the pruned lockfile reports 0 vulnerabilities.
- URL resolution against django 5.2.17: `/service-worker.js` no longer routes and `reverse('ui:service-worker')` raises `NoReverseMatch`, while `/` and `/migrations_notran/` are unaffected. Those two are the only `ui:` names any Python code reverses.
- The CSP `worker-src 'self' blob:` directive stays. It serves the real Web Worker in `src/util/webWorker.js`, which is unrelated.

Existing browsers: a script that fails to evaluate never replaced anything, and nothing was ever registered in a stock deployment, so there is no installed base to clean up. A developer who registered one locally gets it cleared by the dev server's no-op middleware.

Closes #627


cigamit:
Looks like you will need to rebase to resolve the merge conflict.

Aureliolo:
> Looks like you will need to rebase to resolve the merge conflict.
@cigamit 
done :)

cigamit (approved):

PR #658, opened 2026-08-13
Migrate the ascender_devel image and awx/ui build to pnpm

clo-ciq:
## Summary

Part of the org-wide npm/yarn -> pnpm migration tracked in Jira **SRE-1792** (sub-task **SRE-2288**).

Unlike most repos in this migration, `ci.yml` never calls npm/pnpm directly — the `ui-lint`/`ui-test-screens`/`ui-test-general` jobs run `make docker-runner`, which mounts the repo into the `ascender_devel` image and runs `make ui-lint` etc. *inside* the container. So the actual fix lives in the `Makefile` and in `tools/ansible/roles/dockerfile/templates/Dockerfile.j2` (which generates the Dockerfile used to build that image), not in the workflow YAML itself. There is no `actions/setup-node`/`pnpm/action-setup` to add, since Node is installed via a `dnf` module stream inside the Dockerfile, not via a GitHub Action.

- `awx/ui/package.json`: pin `packageManager: pnpm@11.21.0` (with corepack-verified hash).
- Lockfile converted with `pnpm import` (not a fresh install), so resolved dependency versions are unchanged. `package-lock.json` removed.
- **Three phantom dependencies fixed**: `@lingui/core`, `@patternfly/react-styles`, and `eslint-plugin-import-x` were used directly in source/config but only ever present as *transitive* deps of `@lingui/react`, `@patternfly/react-core`, and `eslint-config-airbnb-extended` respectively. npm's flat/hoisted `node_modules` silently allowed this; pnpm's strict, symlinked layout does not, so `eslint` and ~140 Jest suites failed to resolve these modules until they were promoted to direct dependencies (all three resolve to the exact same versions already in the lockfile — no version drift).
- `Makefile`: `NPM_BIN` -> `PNPM_BIN`; every `ui-*` target (`ui-lint`, `ui-test`, `ui-test-screens`, `ui-test-general`, `ui-devel`, `pot`, `po`, etc.) now calls `pnpm`.
- `tools/ansible/roles/dockerfile/templates/Dockerfile.j2`: the `dnf module enable nodejs:22` stream does not bundle corepack (verified by installing it fresh in a throwaway `rockylinux/rockylinux:9` container), so corepack is now bootstrapped explicitly (`npm install -g corepack@0.35.0 && corepack enable`) in both the builder stage and the `ascender_devel` dev-image stage — the two places this template installs Node.
- `awx/ui/Dockerfile`: updated to use `pnpm-lock.yaml` + corepack for consistency, but this Dockerfile is **not referenced by any workflow or docker-compose config** I could find, and was already broken independent of this change (it `ADD`s `.eslintignore`/`.eslintrc.json`/`.linguirc`, which were removed in #391's ESLint-9/flat-config migration). Not verified with a full `docker build` for that reason.
- `docs/development/docker.md`, `.github/copilot-instructions.md`: updated stale npm references.
- No `pnpm-workspace.yaml`/`allowBuilds` added: pnpm blocks postinstall scripts for `esbuild`, `core-js`, `core-js-pure`, and `unrs-resolver`, but I verified none of them are actually needed (esbuild's binary and CLI both work fine as-is; lint/build/test all pass), so I didn't add the file speculatively.

## Test plan

All verified locally (Node v22.23.2, matching the `>=22.22.0` engines constraint and the `nodejs:22` module stream CI uses):

- [x] `pnpm install --frozen-lockfile` succeeds cleanly, no ignored-builds warning
- [x] `pnpm run lint` (eslint) passes
- [x] `pnpm run prettier-check` passes
- [x] `pnpm run pretest && pnpm run test-screens --runInBand` — 376/376 suites, 1918/1918 tests
- [x] `pnpm run pretest && pnpm run test-general --runInBand` — 176/176 suites, 1009/1009 tests
- [x] `pnpm run build` (production webpack build) compiles successfully
- [ ] CI green on the `ui-lint`/`ui-test-screens`/`ui-test-general` matrix jobs in `ci.yml` (depends on the `ascender_devel` image rebuild picking up the corepack change)

Note: `pnpm peers check` surfaces a pre-existing unmet-peer warning (`@babel/core@^8.0.1` vs. several plugins wanting `^7.0.0`) — unrelated to this migration, not blocking any of the above, left as-is.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01EazRi5J66CRzQyVBevEfrW


cigamit:
This is still marked as Draft.  We have changed the packages greatly since this was first proposed, so closing this for now.  We can revisit at a later date.

PR #675, opened 2026-08-18
Let Dependabot open version updates for actions and the UI

blaipr:
## Problem

Dependabot opens security updates here already, #52 and #53 merged in February 2025, #78, #80 and #81 followed in May. Those need no config file. What needs one is ordinary version updates, and there is no `.github/dependabot.yml`, so they never happen.

Everything else has been by hand. The last month holds `Upgrade django to resolve CVE-2026-15307 CVE-2026-15830`, `Upgrade js-yaml`, `Update nanoid`, `Upgrade dompurify`, `Upgrade GitPython` for six advisories, `Upgrade cryptography`, `Upgrade aiohttp`, each of them a bump someone had to notice first.

`ascender-collection` already runs Dependabot for its actions, so the two repos differ for no particular reason.

## What this covers

**GitHub Actions.** 17 `uses:` across five workflows and two composite actions, every one pinned to a SHA with the version in a comment, which Dependabot keeps in step. Composite actions are only picked up when their directory is listed, so both are.

**npm, `/awx/ui`.** 23 dependencies and 70 dev dependencies, the place the manual bumps and the GHSA fixes keep landing. `ui-lint`, `ui-test-screens` and `ui-test-general` already run on every pull request, so a bad bump shows up before review does.

Weekly on Monday, matching the collection. No `groups:` key, so one pull request per dependency. `open-pull-requests-limit: 5` on npm keeps the first weeks readable, and the labels are ones the repo already has, so Dependabot does not have to create any.

Two things sit outside this file: Dependabot bundles dependencies that a lockfile cannot move separately, which is what #80 was, and if grouped security updates are switched on under Settings, that setting wins for security pull requests.

## What this leaves out, and why

**The Python requirements.** `requirements/requirements.txt` is compiled by `requirements/updater.sh`, which then comments out every entry that `requirements_git.txt` installs from git, four of them today including django-ansible-base and ansible-runner. Dependabot compiles the file its own way and knows nothing about that pass, so its output can drift from what `make requirements` produces. Security updates already reach this ecosystem, #81 was `in /requirements`, so the gap is the smallest here and the risk the largest. Worth revisiting deliberately, with the first pull request checked against `updater.sh` output.

**Docker.** `Dockerfile.dev` is on `rockylinux/rockylinux:9` and `awx/ui/Dockerfile` on `node:22-alpine`, both floating majors, so the only bump Dependabot could offer is the major jump this repo is deliberately not taking. The image the product ships from is generated from `tools/ansible/roles/dockerfile/templates/Dockerfile.j2`, which Dependabot cannot read anyway.

**Ignore rules.** None, on purpose. A major React or PatternFly bump is real work rather than a merge, and hiding it in config makes it invisible instead. If those turn out to be noise, an `ignore` entry for `version-update:semver-major` on those two is a one line follow up.

## Tests

Config only, and `.github` is outside what yamllint reads here. Checked by hand:

```
yamllint -s .github/dependabot.yml   # clean
python -c "yaml.safe_load(...)"      # parses, ecosystems: github-actions, npm, no groups key
```


cigamit:
I am debating on this one still.  We have an internal bot we are using that integrates with Snyk that does this, and does the requirements also (stands up a container and runs the updater).  So we may end up with duplicate issue reports.  But Dependabot is sooo slow currently, that it may not matter as we have it patched days before Dependabot finally notices.

For all of these, we still have to manually test each and every one, as I have found issues before uncaught by the automated testing.

blaipr:
Fair point on timing, and the record agrees: #78, #80 and #81 all got closed unmerged.

Those were security updates though, and those already run here without a config file. This only adds version updates, so it cannot duplicate anything Snyk reports. Requirements stays out for exactly the reason you gave: your container running `updater.sh` beats Dependabot's resolution there.

It also lines this repo up with the rest. `ascender-collection` runs Dependabot for actions, `ascender-kit` for actions and pip. `ascender` is the only repo with no config at all.

If the testing burden is the sticking point, happy to cut npm and land the actions half only. That is 17 SHA-pinned refs CI already gates, and it matches what `ascender-collection` runs today.

cigamit (approved):
Can never have too many checks I guess

PR #696, opened 2026-08-23
feat: notify on change

### LINKED ISSUE #206: Notification on Change

### Feature type

Enhancement to Existing Feature

### Feature Summary

We are currently able to create notification on a Job Template that will notify on Started, Successful, or Failure.  We need to add a Changed type, so that we can notify if anything changed during the playbook run.

This is useful for when you run Hardening playbooks in check mode nightly, so you want to notify if anything is out of compliance.

- [x] UI
- [x] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Steps to reproduce

---

### Current results

---

### Sugested feature result

---

### Additional information

_No response_

blaipr:
Closes #206.

## Problem

Notifications can be attached to a job template for start, success and failure. A hardening playbook that runs nightly in check mode succeeds whether or not it found drift, so there is nothing to notify on: the interesting outcome is that something reported as changed.

## Change

Job templates and organizations gain a fourth trigger, **Changed**.

A run that reported a change on any host notifies the templates attached to that trigger, next to the ones for how the run ended. A job that changed something and succeeded notifies both *Success* and *Changed*; one that changed something and failed notifies *Failure* and *Changed*. Check mode counts, which is what makes the trigger useful for a compliance playbook. Start never fires it, since nothing has run yet.

- `notification_templates_changed` on the job template and on the organization, with the endpoints and the related links that the other triggers have.
- `Job.has_changes()` is true when any of the job's host summaries recorded a change. Ad hoc commands answer the same way, so an organization's changed templates cover them as they cover its jobs. Project updates, inventory updates, system jobs and workflow jobs record no per host results and never fire it.
- The trigger only costs a query when changed templates are actually attached: the dispatch checks for them before asking whether the run changed anything.
- The message is customizable like the other triggers, and defaults to `{{ job_friendly_name }} #{{ job.id }} '{{ job.name }}' reported changes: {{ url }}`.

In the UI the Notifications tab of a job template and of an organization gains a Changed toggle, and the notification template form gains the Changed message fields.

Workflow job templates are left out: a workflow does not run against hosts itself, and the job templates inside it each notify on their own.

## Testing

- New model tests cover: a job whose host summaries recorded no change, one that recorded a change, an ad hoc command that recorded one, the job types that never report changes, a changed job notifying both the outcome and the changed templates, a job that changed nothing notifying only the outcome, a failed job that changed something notifying both, start not firing the trigger, and an unknown status still being refused.
- New API tests cover the two endpoints, plus that the job template aggregates its own changed templates and the ones on its organization.
- New UI tests cover the Changed toggle rendering, its two click paths, and the list fetching and disassociating through it.
- `flake8 awx` and `black --check awx` pass, and prettier passes on every changed file under `awx/ui`.

The suites themselves were not run locally, since that needs the development image; CI runs them on this pull request.


blaipr:
Thanks, this review earned its keep. Both findings were right, and chasing the first one turned up a real bug in the code rather than in the test. Fixed in 0e9034356d.

**`has_changes()` on `AdHocCommand` was broken.** The parametrized test flagged above was failing, and once I ran the class properly a third failure came out with it:

```
FAILED test_job_types_without_host_results_never_report_changes[InventoryUpdate]
FAILED test_job_types_without_host_results_never_report_changes[ProjectUpdate]
FAILED test_ad_hoc_command_with_changes
AttributeError: 'AdHocCommand' object has no attribute 'job_host_summaries'
```

That last one is not a test problem. I had written `AdHocCommand.has_changes()` as a copy of `Job.has_changes()`, and an ad hoc command has no `JobHostSummary` rows: its per host results are `AdHocCommandEvent` records, related as `ad_hoc_command_events`, and the `changed` flag there is a boolean rather than a count. So the method raised `AttributeError` for every ad hoc command, which `send_notification_templates` would have hit in production the moment one finished.

```python
-        return self.job_host_summaries.filter(changed__gt=0).exists()
+        return self.ad_hoc_command_events.filter(changed=True).exists()
```

I also added the negative case, `test_ad_hoc_command_without_changes`, since only the positive one existed and neither was really running.

**The parametrized test.** Correct on both counts: `InventoryUpdate` and `ProjectUpdate` cannot be created from a name alone. The first thing that actually stops it is `NOT NULL constraint failed: main_inventoryupdate.source`, before the `inventory_source` foreign key even comes up. Rather than invent an inventory source and a project to reach a method that never queries anything, I dropped the database from the test:

```python
-    @pytest.mark.django_db
     @pytest.mark.parametrize('JobClass', [InventoryUpdate, ProjectUpdate, SystemJob, WorkflowJob])
     def test_job_types_without_host_results_never_report_changes(self, JobClass):
-        assert JobClass.objects.create(name='fake-job').has_changes() is False
+        assert JobClass().has_changes() is False
```

These four inherit `JobNotificationMixin.has_changes()`, which returns `False` without touching the database, so an unsaved instance tests the contract exactly and the test stops depending on relations that have nothing to do with what it asserts. If one of them ever gains a real implementation, this fails loudly, which is what you want from it.

**The two docstrings.** Updated both to `"approvals"`, `"started"`, `"success"`, `"error"` and `"changed"`. Worth noting they were already stale before this pull request: `approvals` and `started` have been handled by both helpers for a while and were never documented. Now the docstring matches the branches in the function.

Tests after the change:

```
py.test awx/main/tests/functional/models/test_notifications.py \
        awx/main/tests/functional/api/test_notifications.py
49 passed in 39.01s
```

`black --check` and `flake8` are clean on the touched Python files.

The whole suite, on a freshly created database, in a checkout linked with `make awx-link`:

```
py.test -p no:cacheprovider --create-db -n auto --dist=loadfile \

1 failed, 3835 passed, 10 skipped in 403.72s (0:06:43)
```

That failure is unrelated: `test_proxy_ip_allowed` patches `REMOTE_HOST_HEADERS` and `PROXY_IP_ALLOWED_LIST` on the settings singleton and intermittently sees another worker's state under `-n auto`. It is the flake instrumented in #694 and it passes solo. `main` at `e197d7cc0f` reports 3817 passed on the same command, so this branch adds 18 tests and no failures.

cigamit:
After this PR, it was wanting to create a new 210 migration (and ui tests are failing), so I ran this through Fable internally, and it came up with these.

### Issues found
1) **Migration state mismatch** — the models declare _related_name='%(class)s_notification_templates_for_changed'_, but **0209_notification_templates_changed.py** hard-coded the resolved names (_jobtemplate_..., organization_..._). Django's autodetector compares these literally, so _makemigrations --check_ failed and wanted a corrective 0210 _AlterField_ migration. I changed 0209 to use the _%(class)s_ form. Since related_name is ORM-only metadata, this doesn't touch the DB schema — the migration you already ran is still valid, no re-run needed. makemigrations --check is now clean.

2) **10 pre-existing UI tests broken** — **NotificationTemplateForm.js:161** now unconditionally reads _defs.changed_, but the fixture in **NotificationTemplateForm.test.js** didn't include a changed key, crashing the form render. I added _changed: messageDef_ to the fixture (the real prop comes from server OPTIONS, which now always includes it — the form already hard-depends on the other keys the same way, so hardening the code instead would have been off-convention).

3) **Model default left behind** — NotificationTemplate.default_messages() at **notifications.py** still returned only started/success/error/workflow_approval, so templates created without explicit messages exposed a messages dict missing the changed key. Harmless functionally (the send path falls back via _.get('changed', {}))_, but inconsistent in API responses. Added _'changed':_ None plus the matching assertion in _test_basic_parameterization_. No migration impact (callable default).

```
diff --git a/awx/main/migrations/0209_notification_templates_changed.py b/awx/main/migrations/0209_notification_templates_changed.py
index 21bf8505..838a1301 100644
--- a/awx/main/migrations/0209_notification_templates_changed.py
+++ b/awx/main/migrations/0209_notification_templates_changed.py
@@ -10,11 +10,11 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='jobtemplate',
             name='notification_templates_changed',
-            field=models.ManyToManyField(blank=True, related_name='jobtemplate_notification_templates_for_changed', to='main.notificationtemplate'),
+            field=models.ManyToManyField(blank=True, related_name='%(class)s_notification_templates_for_changed', to='main.notificationtemplate'),
         ),
         migrations.AddField(
             model_name='organization',
             name='notification_templates_changed',
-            field=models.ManyToManyField(blank=True, related_name='organization_notification_templates_for_changed', to='main.notificationtemplate'),
+            field=models.ManyToManyField(blank=True, related_name='%(class)s_notification_templates_for_changed', to='main.notificationtemplate'),
         ),
     ]
diff --git a/awx/main/models/notifications.py b/awx/main/models/notifications.py
index 70eaf7c3..aa5cfce3 100644
--- a/awx/main/models/notifications.py
+++ b/awx/main/models/notifications.py
@@ -73,7 +73,7 @@ class NotificationTemplate(CommonModelNameNotUnique):
     notification_configuration = prevent_search(models.JSONField(default=dict))
 
     def default_messages():
-        return {'started': None, 'success': None, 'error': None, 'workflow_approval': None}
+        return {'started': None, 'success': None, 'error': None, 'changed': None, 'workflow_approval': None}
 
     messages = models.JSONField(null=True, blank=True, default=default_messages, help_text=_('Optional custom messages for notification template.'))
 
diff --git a/awx/main/tests/functional/test_notifications.py b/awx/main/tests/functional/test_notifications.py
index cf93030a..c635bc57 100644
--- a/awx/main/tests/functional/test_notifications.py
+++ b/awx/main/tests/functional/test_notifications.py
@@ -47,7 +47,7 @@ def test_basic_parameterization(get, post, user, organization):
     assert 'url' in response.data['notification_configuration']
     assert 'headers' in response.data['notification_configuration']
     assert 'messages' in response.data
-    assert response.data['messages'] == {'started': None, 'success': None, 'error': None, 'workflow_approval': None}
+    assert response.data['messages'] == {'started': None, 'success': None, 'error': None, 'changed': None, 'workflow_approval': None}
 
 
 @pytest.mark.django_db
diff --git a/awx/ui/src/screens/NotificationTemplate/shared/NotificationTemplateForm.test.js b/awx/ui/src/screens/NotificationTemplate/shared/NotificationTemplateForm.test.js
index d293d335..46ced336 100644
--- a/awx/ui/src/screens/NotificationTemplate/shared/NotificationTemplateForm.test.js
+++ b/awx/ui/src/screens/NotificationTemplate/shared/NotificationTemplateForm.test.js
@@ -66,6 +66,7 @@ const defaults = {
   started: messageDef,
   success: messageDef,
   error: messageDef,
+  changed: messageDef,
   workflow_approval: {
     approved: messageDef,
     denied: messageDef,

```

blaipr:
Thanks, all three were right. Applied in 5dbfcda493, and the branch now carries current `main`.

**Migration state mismatch.** Confirmed and fixed. I reproduced it before and after: with the hard-coded names, `check_migrations` wants a corrective 0210 and exits 1.

```
Migrations for 'main':
  awx/main/migrations/0210_missing_migration_file.py
    ~ Alter field notification_templates_changed on jobtemplate
    ~ Alter field notification_templates_changed on organization
```

With the `%(class)s` form it is `No changes detected`. That is also the convention the rest of the tree already uses: 0184 writes `%(class)s_notification_templates_for_errors` and `%(class)s_notification_templates_for_success` for the same relations. Worth noting `api-migrations` stayed green through this, since `test_migrations` never runs the autodetector check, so it only surfaces via `make test`.

**The 10 UI tests.** Same call, and for the same reason: `NotificationTemplateForm` reads `defs.changed` exactly the way it reads `defs.started`, so the fixture is what was stale, not the code. `NotificationTemplateForm.test.js` is 10 passed now, and the wider notification UI is 8 suites, 55 passed.

**`default_messages()`.** Added `'changed': None` and the matching assertion. Agreed it is cosmetic given the `.get('changed', {})` fallback, but the OPTIONS response and the stored default should not disagree.

Verification after the merge:

```
awx-manage check_migrations --dry-run --check -n missing_migration_file
No changes detected

py.test awx/main/tests/functional/test_notifications.py \
        awx/main/tests/functional/models/test_notifications.py \
        awx/main/tests/functional/api/test_notifications.py
60 passed in 38.66s

npx jest src/components/NotificationList src/screens/NotificationTemplate
Test Suites: 8 passed, Tests: 55 passed
```

`black --check` and `flake8` are clean on the touched Python files, and prettier is clean on the touched JS.

cigamit:
Main has a new migration file, so will need to bump this one up to 210 now, then it should be good to merge.

blaipr:
Renumbered to `0210_notification_templates_changed`, depending on `0209_cleanup_dab_rbac_leftovers`, and the branch carries current `main`.

```diff
 class Migration(migrations.Migration):
     dependencies = [
-        ('main', '0208_alter_job_skip_tags_alter_jobtemplate_skip_tags'),
+        ('main', '0209_cleanup_dab_rbac_leftovers'),
     ]
```

The graph is a single leaf again:

```
leaf nodes for main: [('main', '0210_notification_templates_changed')]
plan tail: 0208_alter_job_skip_tags... -> 0209_cleanup_dab_rbac_leftovers -> 0210_notification_templates_changed
```

The rename carries no schema change, so anyone who already ran the file as 0209 needs nothing beyond the row in `django_migrations` naming the new one.

The merge itself was clean. `main` touched four of the files this branch touches, all of them the `ansible_base.lib.utils.models` to `awx.dab.lib.utils.models` move plus the DRF 3.16 `required`/`default` work in `serializers.py`, none of it overlapping the changed trigger. The `'changed': None` in `default_messages()` and the `has_changes()` implementations all came through intact.

cigamit (approved):

PR #718, opened 2026-08-23
build: bump drf-spectacular from 0.29.0 to 0.30.0 in /requirements

blaipr:
##### SUMMARY

Bumps `drf-spectacular` from `0.29.0` to `0.30.0` in `requirements/requirements.txt`. It generates the OpenAPI schema the API serves.

```
requirements/updater.sh upgrade drf-spectacular
```

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ASCENDER VERSION

```
25.5.1
```

## Tests

Tested in the same image, with `drf-spectacular 0.30.0` installed into the AWX venv:

```
  1403 passed, 1 skipped in 11.41s

py.test awx/main/tests/functional/api/test_generic.py awx/main/tests/functional/api/test_settings.py
  46 passed in 18.35s
```

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -149,7 +149,7 @@ djangorestframework==3.15.2
     #   drf-spectacular
 djangorestframework-yaml==2.0.0
     # via -r /awx_devel/requirements/requirements.in
-drf-spectacular==0.29.0
+drf-spectacular==0.30.0
     # via -r /awx_devel/requirements/requirements.in
 durationpy==0.10
     # via kubernetes


cigamit (dismissed):

blaipr:
Rebased onto current `main` to clear the conflict. Still the same one line, `drf-spectacular==0.29.0` to `0.30.0`; the conflict was only the neighbouring pins that landed in the meantime.

Heads up that the force push dismissed your approval, so this needs another one when you get a chance.

cigamit (approved):

PR #719, opened 2026-08-23
build: bump python-daemon from 3.0.1 to 3.1.2 in /requirements

blaipr:
##### SUMMARY

Bumps `python-daemon` from `3.0.1` to `3.1.2` in `requirements/requirements.txt`. It daemonizes the dispatcher and the callback receiver.

```
requirements/updater.sh upgrade python-daemon
```

run inside the `ascender_devel` image, which is where that script insists on running.

`docutils` drops out of the lockfile with it: python-daemon 3.0.1 depended on it and 3.1.2 does not, and nothing else in the tree pulls it. The docs build keeps its own copy in `docs/docsite/requirements.txt`, which is untouched here. The result is 1 added / 4 removed in the lockfile.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ASCENDER VERSION

```
25.5.1
```

## Tests

Tested in the same image, with `python-daemon 3.1.2` installed into the AWX venv:

```
  1403 passed, 1 skipped in 11.35s

py.test awx/main/tests/unit/test_tasks.py
  136 passed in 3.82s
```

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -149,8 +149,6 @@ djangorestframework==3.15.2
     #   drf-spectacular
 djangorestframework-yaml==2.0.0
     # via -r /awx_devel/requirements/requirements.in
-docutils==0.20.1
-    # via python-daemon
 drf-spectacular==0.29.0
     # via -r /awx_devel/requirements/requirements.in
 durationpy==0.10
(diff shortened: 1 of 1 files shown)


cigamit:
We can upgrade this, as we only include it in for ansible-runner's sake (we don't use it ourselves).  We may look at forking ansible-runner so we can remove this dependency completely (and start fixing other issues).

cigamit (approved):

PR #726, opened 2026-08-23
build: bump cachetools from 5.3.2 to 7.1.7 in /requirements

blaipr:
##### SUMMARY

Bumps `cachetools` from `5.3.2` to `7.1.7` in `requirements/requirements.txt`. It is pulled in as a direct requirement and used by the Google auth stack under it.

```
requirements/updater.sh upgrade cachetools
```

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ASCENDER VERSION

```
25.5.1
```

## Tests

Tested in the same image, with `cachetools 7.1.7` installed into the AWX venv:

```
  1403 passed, 1 skipped in 11.91s

py.test awx/main/tests/unit/test_capacity.py awx/main/tests/functional/api/test_settings.py
  61 passed in 14.93s
```

It crosses two majors, so the whole suite ran too, on a freshly created database:

```
py.test --create-db -n auto --dist=loadfile \

  1 failed, 3815 passed, 10 skipped in 7m51s
```

That failure is not this bump. `test_proxy_ip_allowed` patches `REMOTE_HOST_HEADERS` and `PROXY_IP_ALLOWED_LIST` on the settings singleton, and under `-n auto` it intermittently sees another worker's settings state. With cachetools 7.1.7 installed it passes solo (7 passed) and passes alongside the settings tests it shares a worker with (46 passed), and I saw the same failure once during an unrelated bump on the same day. The serial baseline for this tree is 3816 passed, 10 skipped, with no failures.

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -54,7 +54,7 @@ botocore==1.43.78
     #   -r /awx_devel/requirements/requirements.in
     #   boto3
     #   s3transfer
-cachetools==5.3.2
+cachetools==7.1.7
     # via -r /awx_devel/requirements/requirements.in
 # certifi @ git+https://github.com/ansible/system-certifi.git@devel  # git requirements installed separately
     # via


blaipr:
Rebased onto current `main` so this sits on top of the bumps that landed since. Still the same one line, `cachetools==5.3.2` to `7.1.7`.

cigamit (approved):

PR #749, opened 2026-08-26
build: bump receptorctl from 1.6.2 to 1.6.7 in /requirements

blaipr:
##### SUMMARY

Bumps `receptorctl` from `1.6.2` to `1.6.7` in `requirements/requirements.txt`. It is the client the task system uses to talk to the receptor mesh.

```
requirements/updater.sh upgrade receptorctl
```

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ASCENDER VERSION

```
25.5.1
```

## Tests

Tested before opening, in the same image, with `receptorctl 1.6.7` installed into the AWX venv:

```
  1404 passed, 1 skipped in 10.65s
```

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -362,7 +362,7 @@ pyyaml==6.0.3
     #   drf-spectacular
     #   kubernetes
     #   receptorctl
-receptorctl==1.6.2
+receptorctl==1.6.4
     # via -r /awx_devel/requirements/requirements.in
 referencing==0.33.0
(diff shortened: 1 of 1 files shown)


blaipr:
Closing in favour of a correctly-titled pull request: the compile resolves receptorctl to 1.6.4, not 1.6.7, and this pull request's title and test run both claimed 1.6.7. Reopened as the 1.6.4 bump with the cap explained.

PR #752, opened 2026-08-26
build: bump packaging from 24.2 to 26.3 in /requirements

blaipr:
##### SUMMARY

Bumps `django-guid` from `3.5.2` to `3.6.1` in `requirements/requirements.txt`. It stamps the correlation ID that ties a request's log lines together.

```
# requirements.in:  django-guid==3.5.2  ->  django-guid==3.6.1
requirements/updater.sh run
```

run inside the `ascender_devel` image, which is where that script insists on running. This one is pinned with `==` in `requirements.in`, so `upgrade` cannot move it: the pin is edited first and `run` recompiles against it. The result is 1 added / 1 removed in each of the two files.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ASCENDER VERSION

```
25.5.1
```

## Tests

Tested before opening, in the same image, with `django-guid 3.6.1` installed into the AWX venv:

```
  1404 passed, 1 skipped in 10.36s
```

### DIFF
--- a/docs/docsite/requirements.txt
+++ b/docs/docsite/requirements.txt
@@ -23,7 +23,7 @@ jinja2==3.1.6
     #   sphinx
 markupsafe==3.0.3
     # via jinja2
-packaging==26.0
+packaging==26.3
     # via sphinx
 pygments==2.20.0
     # via


blaipr:
Closing: opened by mistake against a stale branch. `build/bump-packaging-26.3` already existed on the fork from #701, the docs lockfile bump that has since merged, so this pull request carries that merged `docs/docsite/requirements.txt` change rather than the production lockfile one its title describes. Reopening on a distinct branch.

PR #755, opened 2026-08-26
build: bump wheel from 0.46.3 to 0.48.0 in /requirements

blaipr:
##### SUMMARY

Bumps `wheel` from `0.46.3` to `0.48.0` in `requirements/requirements.txt`. It builds the wheels the image installs.

```
requirements/updater.sh upgrade wheel
```

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ASCENDER VERSION

```
25.5.1
```

## Tests

Tested before opening, in the same image, with `wheel 0.48.0` installed into the AWX venv:

```
  1404 passed, 1 skipped in 11.85s
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ RECEPTOR_IMAGE ?= quay.io/ansible/receptor:devel
 SRC_ONLY_PKGS ?= cffi,pycparser,psycopg,twilio
 # These should be upgraded in the AWX and Ansible venv before attempting
 # to install the actual requirements
-VENV_BOOTSTRAP ?= pip==26.2.1 setuptools==83.0.0 setuptools_scm[toml]==9.2.2 wheel==0.46.2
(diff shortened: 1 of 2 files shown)


cigamit:
`wheel` should always be bumped simultaneously in the Makefile also.

blaipr:
Right, and it had already drifted: the bootstrap was on `wheel==0.46.2` while `requirements.txt` was on `0.46.3`. Both are on `0.48.0` now.

```diff
-VENV_BOOTSTRAP ?= pip==26.2.1 setuptools==83.0.0 setuptools_scm[toml]==9.2.2 wheel==0.46.2
+VENV_BOOTSTRAP ?= pip==26.2.1 setuptools==83.0.0 setuptools_scm[toml]==9.2.2 wheel==0.48.0
```

`requirements/updater.sh` pins `pip` and `setuptools` but not `wheel`, so there is no third place for this one. That is the setuptools bump in #759, which does carry it.

Also rebased onto current `main`, which clears the `licenses/docutils.txt` failure `test_python_and_js_licenses` was reporting on the older base.

cigamit (approved):

PR #759, opened 2026-08-26
build: bump setuptools from 83.0.0 to 84.0.0 in /requirements

blaipr:
##### SUMMARY

Bumps `setuptools` from `83.0.0` to `84.0.0` in `requirements/requirements.txt`. It is the build backend the image uses for every source distribution it installs.

```
# requirements.in:  setuptools==83.0.0  ->  setuptools==84.0.0   (comment kept)
requirements/updater.sh run
```

run inside the `ascender_devel` image, which is where that script insists on running. Pinned with `==` in `requirements.in`, so `upgrade` cannot move it: the pin is edited first and `run` recompiles against it. The result is 1 added / 1 removed in each of the two files.

The `# CVE-2026-59890` comment is deliberately kept on the pin. 83.0.0 is where that fix landed, so 84.0.0 stays above it and the reason for pinning still reads true; the comment would only be wrong if the pin ever moved down.

Like the other build-toolchain bumps in this batch, the unit suite passing says less here than it does for a library: setuptools is exercised when the image builds a source distribution, not by the tests. The real check is a green image build, which happens once the workflow is approved on this fork pull request.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ASCENDER VERSION

```
25.5.1
```

## Tests

Tested before opening, in the same image, with `setuptools 84.0.0` installed into the AWX venv:

```
  1404 passed, 1 skipped in 12.84s
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ RECEPTOR_IMAGE ?= quay.io/ansible/receptor:devel
 SRC_ONLY_PKGS ?= cffi,pycparser,psycopg,twilio
 # These should be upgraded in the AWX and Ansible venv before attempting
 # to install the actual requirements
-VENV_BOOTSTRAP ?= pip==26.2.1 setuptools==83.0.0 setuptools_scm[toml]==9.2.2 wheel==0.48.0
+VENV_BOOTSTRAP ?= pip==26.2.1 setuptools==84.0.0 setuptools_scm[toml]==9.2.2 wheel==0.48.0
 
 NAME ?= awx
 
--- a/requirements/requirements.in
+++ b/requirements/requirements.in
@@ -71,7 +71,7 @@ uWSGI>=2.0.22 # CVE-2023-27522
 valkey[libvalkey]
 wheel>=0.46.2  # CVE-2026-24049
 pip==26.2.1  # CVE-2026-13346
-setuptools==83.0.0  # CVE-2026-59890
+setuptools==84.0.0  # CVE-2026-59890
 setuptools-scm[toml]  # see UPGRADE BLOCKERs, xmlsec build dep
 setuptools-rust>=0.11.4  # cryptography build dep
 pkgconfig>=1.5.1  # xmlsec build dep - needed for offline build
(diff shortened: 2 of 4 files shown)


cigamit:
`setuptools` should always be bumped simultaneously in the Makefile and the requirements/updater.sh also.

blaipr:
Done, both of them:

```diff
-VENV_BOOTSTRAP ?= pip==26.2.1 setuptools==83.0.0 setuptools_scm[toml]==9.2.2 wheel==0.46.2
+VENV_BOOTSTRAP ?= pip==26.2.1 setuptools==84.0.0 setuptools_scm[toml]==9.2.2 wheel==0.46.2
```

```diff
   # pip / setuptools version must match the version used in AWX venv (see README.md UPGRADE BLOCKERs)
-  "${venv}/bin/python3" -m pip install -U 'pip==26.2.1' 'setuptools==83.0.0' pip-tools
+  "${venv}/bin/python3" -m pip install -U 'pip==26.2.1' 'setuptools==84.0.0' pip-tools
```

Those are the only two `setuptools==` pins outside `requirements/`, so the four places that name a version now agree: `requirements.in`, `requirements.txt`, the Makefile bootstrap and the updater.

Also rebased onto current `main`, which clears the `licenses/docutils.txt` failure `test_python_and_js_licenses` was reporting on the older base.

cigamit (dismissed):

cigamit:
More conflicts thanks to #755

blaipr:
Rebased onto current `main`, conflict is gone. The collision was the `VENV_BOOTSTRAP` line, since #755 moved `wheel` on it while this branch still carried the old value:

```diff
-VENV_BOOTSTRAP ?= pip==26.2.1 setuptools==83.0.0 setuptools_scm[toml]==9.2.2 wheel==0.48.0
+VENV_BOOTSTRAP ?= pip==26.2.1 setuptools==84.0.0 setuptools_scm[toml]==9.2.2 wheel==0.48.0
```

`wheel==0.48.0` from #755 stays, `setuptools==84.0.0` is the one this branch changes. The other three files merged cleanly, so the diff against `main` is again only the four places that name a `setuptools` version: `requirements.in`, `requirements.txt`, the Makefile bootstrap and `requirements/updater.sh`.

cigamit (approved):

PR #760, opened 2026-08-26
fix: make the workflow job selector show and switch the right job

### LINKED ISSUE #742: Unable to switch different workflow task using upperright workflow-task-selector

### Bug Summary

When viewing details of a task that ran as part of a workflow, a drop-down is displayed in the upper right corner showing "Workflowtask 1/X", "X" being the total number of tasks in the parenting workflow. 
This drop-down seems to be mainly defective, not updating the current task view, nor showing the current task sequence id.

### Ascender version

25.5.1

- [x] UI
- [ ] UI (tech preview)
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

kubernetes

### Modifications

no

### Ansible version

2.16

### Operating system

openSUSE

### Web browser

Firefox

### Steps to reproduce

- Create and run a workflow containing at least 2 tasks
- Open detail view of one of the tasks while or after running the workflow
- Notice the upper-right dropdown showing `Workflowtask 1/X`, `X` being the total number of tasks in the workflow. But also notice it also shows current task as being task `1`, even if you open details for the second or any subsequent tasks.
- Click on the drop-down and select another task from the workflow

### Expected results

- The drop-down shows `Workflowtask 2/X` for the second task, `3/X` for the third, and so on..
- When clicking on another task within the drop-down, the actual task detail view of that task is displayed. 

### Actual results

- The drop-down always shows `Workflowtask 1/X` even for the second task or any subsequent tasks
- When clicking on another task within the drop-down, nothing happens

### Additional information

_No response_

blaipr:
##### SUMMARY

Fixes the workflow job selector in the job tab bar. Closes #742.

The report is two symptoms, and they turned out to have three separate causes, only one of them in the selector's own code.

**The counter was the literal `1`.** `WorkflowOutputNavigation` rendered ``t`Workflow Job 1/${relevantResults.length}` ``, and `relevantResults` filtered out the job you were looking at, so the total was short by one too. A four task workflow read `1/3` on every one of its tasks. It now counts every node that ran a job, the current one included, and reports where in that list the job on screen sits.

**Picking a task did nothing, and that is not a bug in the selector.** The selector is registered in `Job.js` as a tab with `link: undefined`, so that it can sit in the tab bar:

```js
tabsArray.push({
  name: <WorkflowOutputNavigation parentRef={ref} relatedJobs={relatedJobs} />,
  link: undefined,
  id: 2,
  hasstyle: 'margin-left: auto',
});
```

A click inside it bubbles up to the PatternFly tab, `RoutedTabs.handleTabSelect` matched tab id 2, and ran `navigate(undefined)`. That resolves to the current url, so it landed you back where you were and undid the navigation the selector had just performed. `RoutedTabs` now leaves a tab with no link alone.

**`href={!tab.hasstyle && `#${tab.link}`}` is `false` for that tab**, and React rejects `false` as an href value. It was logging `Received 'false' for a non-boolean attribute 'href'` on every job page.

Two more came out of writing the tests:

- With a status filter applied, the toggle rendered a `Label` with `onClose`, which is a close `<button>` inside the toggle's own `<button>`. Invalid HTML, and that close button was unclickable for the same bubbling reason. The filter is cleared by picking the same status again.
- The menu seeded `useState` from the first render, so after moving between jobs it still offered the list belonging to the page you came from.

`ad_hoc_command` was also missing from the type to url-segment map, which would have built `/jobs/undefined/<id>/output` for an ad hoc command node.

##### ISSUE TYPE

##### COMPONENT NAME
 - UI

##### ASCENDER VERSION

```
25.5.1
```

##### ADDITIONAL INFORMATION

**On the tests.** The existing `WorkflowOutputNavigation.test.js` mocked out the component under test and asserted against the mock, so it exercised none of this code and passed throughout. It is replaced by twelve tests against the real component, plus two on `RoutedTabs` for tabs that host a control.

Worth knowing for review: **jsdom passes the navigation case that a real browser fails.** The bug lives in event bubbling through nested buttons, which jsdom does not model faithfully, so the jest test for "picking a job navigates" went green against the broken code. I only found the real cause by driving a headless browser against a running instance. That is the argument for end-to-end coverage, raised in #761 and built in #762, whose regression specs are cover for exactly this bug: they fail five ways against this branch's parent and pass once it lands.

**Verified against a running instance**, workflow job 951, four job nodes:

```
counter on job 952: "Workflow Job 1/4"
counter on job 953: "Workflow Job 2/4"
counter on job 954: "Workflow Job 3/4"
counter on job 955: "Workflow Job 4/4"
navigate 1 -> node3:            #/jobs/playbook/954/output
counter follows:                "Workflow Job 3/4"
navigate again without reload:  #/jobs/playbook/953/output
filter chip shown, all 4 still listed under Successful
console errors during interaction: none
```

`RoutedTabs` is used by 46 screens, so tab navigation was checked in the browser too, and none of it changes:

```
PASS  job: Output -> Details          PASS  workflow job: Output -> Details
PASS  job: Details -> Output          PASS  workflow job: Details -> Output
PASS  templates list loads            PASS  job: Back to Jobs (persistentFilterKey path)
```

Full UI suite after the change:

```
npm --prefix awx/ui run test -- --watchAll=false
  Test Suites: 552 passed, 552 total
  Tests:       2949 passed, 2949 total

npm --prefix awx/ui run lint
  clean
```

**Not fixed here, and worth a follow-up.** Rendering a control inside a tab's `<button>` is invalid HTML whatever else is done, and it is the reason these clicks misbehaved in the first place. Only `Job.js` does it, one `hasstyle` usage in the whole tree, so moving that control out of the tab list would remove the whole class rather than guarding each symptom. I kept this change to the reported bug.


cigamit:
Seems to be a linter issue.  The other 2 failures weren't caused by this PR, but some earlier ones.

The only other thing I notice is that when selecting a filter (click failed) the menu shrinks to the size of the input box, which itself shrunk.  This makes some of the text truncate off the screen.  We should probably set that input box to minimum width.

image

blaipr:
Both fixed, and you were right about the linter.

**The lint.** Prettier, on the test file only:

```
[warn] src/components/WorkflowOutputNavigation/WorkflowOutputNavigation.test.js
[warn] Code style issues found in the above file.
```

Formatting alone, no assertions changed. `eslint .` and `prettier --check` are clean across the whole of `src/` now.

**The shrinking menu.** Same cause as the truncation, and it is not the input box: PatternFly's `Popper` defaults to `minWidth: 'trigger'`, so the menu's minimum width is whatever the toggle is currently rendering at.

```js
const Popper = ({ ..., minWidth = 'trigger', ... }) => {
  ...
  state.styles.popper.minWidth = minWidth === 'trigger' ? `${triggerWidth}px` : minWidth;
```

Picking a status swaps `Workflow Job 3/12` in the toggle for a single short chip, the toggle shrinks to fit it, and the menu follows it down until the only thing holding it open is the filter input, which is what your screenshot shows. So the floor goes on the toggle, and the menu inherits it:

```js
const WorkflowMenuToggle = styled(MenuToggle)`
  min-width: 220px;
`;
```

220px leaves room for the longer translations of the position text. Matches how `AdvancedSearch` handles the same thing.

There is a regression test for it, which fails on `Expected: "220px" / Received: ""` without the change:

```js
test('keeps a width floor on the toggle once a status filter is on', ...)
```

**Verification.**

```
npx jest src/components/WorkflowOutputNavigation src/components/RoutedTabs
  2 suites, 19 passed

npx jest --testPathIgnorePatterns='<rootDir>/src/screens/'
  176 suites, 1024 passed
```

The wider run matters here because `RoutedTabs` is used by a lot of screens.

**The other two failures.** Agreed, neither is this branch. `api-test` was `test_python_and_js_licenses` failing on `licenses/docutils.txt`, which `main` has since removed, and `dev-env` timed out at the hour. The branch is rebased onto current `main` now, which also clears the conflict it had picked up, so both should come back on their own.

The locale diff is much smaller than it was: rebasing onto current `main` means extraction only rewrites the two new messages and the line references that actually moved, rather than every catalog end to end.

cigamit (approved):

Issue #761, opened 2026-08-26
Add an end-to-end browser test suite for the UI


blaipr:
### Please confirm the following

- [x] I have checked the [current issues](https://github.com/ctrliq/ascender/issues) for duplicates.
- [x] I understand that Ascender is open source software provided for free and that I might not receive a timely response.

### Feature type

New Feature

### Feature Summary

A small end-to-end browser suite for the UI, implemented in #762 and running as a CI job. The unit suite is large and healthy, 552 files and 2949 tests, but it runs entirely in jsdom, and there is a class of defect it cannot see.

#742 is the case that prompted this. The workflow job selector never switched jobs. Writing an ordinary jest test for "picking a job navigates" produced a **passing test against the broken code**, because the cause was a click bubbling from a nested `<button>` into the PatternFly tab that hosts the control, and jsdom does not model that faithfully. The same interaction in a real browser did nothing at all, which is what users reported. The fix is #760; the point here is that the unit suite could not have caught it, and adding more unit tests would not change that.

Coverage is not the gap. It currently sits at 75.6% of statements and 71.6% of branches, and the file at the centre of #742 had a test all along: that test mocked out the component under test and asserted against the mock, so it passed for as long as the bug existed while contributing coverage the whole time.

### What it covers

Not everything, deliberately. A handful of flows where a break is both likely and expensive, each asserting on the url and on rendered text rather than on internals.

**In the suite, delivered by #762:**

- Log in, and the redirect an unauthenticated visit gets
- Job tab navigation, which around forty screens share through `RoutedTabs`
- The workflow job selector: position, switching, a second switch without a page load, filtering

**Candidates for later, not built yet:**

- Launch a job template and watch the output reach a terminal state
- Create, edit and delete one resource, an inventory or a project
- A schedule and a notification template round trip, since both have had form regressions before

### Why a browser, specifically

The failures this catches are the ones jsdom cannot represent: event bubbling through invalid nesting, real focus and keyboard behaviour, PatternFly menus and poppers that render through portals, and anything depending on layout. #742 was the first kind. The nested `<button>` that caused it also logs `In HTML, <button> cannot be a descendant of <button>` on every job page, which no test currently notices.

### How it is built

- Playwright, headless, one project rather than a matrix.
- It runs against a stack that is already up, the one `make docker-compose` gives you, so it needs no new infrastructure to be useful locally.
[...]

PR #762, opened 2026-08-26
test: add an end-to-end browser suite for the UI

blaipr:
##### SUMMARY

Adds an end-to-end browser suite for the UI and runs it as a CI job. Related #761.

**Stacked on #760.** This branch carries that fix as well, because the suite's regression specs are cover for exactly that bug and would fail without it. Once #760 merges this reduces to `awx/ui/e2e/`, the CI job, and one line in the eslint config.

##### Why a browser and not more jest

The jest suite is large and healthy, 552 files and around 2950 tests at roughly 75% of statements. It runs in jsdom, which does not reproduce event bubbling through the DOM, focus, or anything rendered through a portal.

#742 is the worked example. A jest test asserting "picking a job navigates" **passed against code where the click did nothing in every real browser**: the click bubbled out of the selector into the tab that hosts it, and `RoutedTabs` called `navigate(undefined)`, landing back on the current url. No amount of unit testing reaches that.

Coverage was not the gap either. The file at the centre of #742 had a test all along, one that mocked out the component under test and asserted against the mock. It passed for as long as the bug existed.

##### What it covers

Ten specs, deliberately few, over flows where a break is likely and expensive:

| File | Covers |
| ---- | ------ |
| `login.spec.js` | signing in, and that an unauthenticated visit is redirected |
| `tabs.spec.js` | `RoutedTabs`, which around forty screens share |
| `job-output.spec.js` | the workflow job selector: position, switching, a second switch without a page load, filtering |

They assert on the url and on rendered text rather than on internals, and fail on console errors the page is not expected to log. Two of the three defects in #742 first announced themselves that way.

##### The fixtures

`seed.js` runs once before the specs and builds a workflow from **system job templates**. Every instance has them, they need no project, inventory, credential or network access, and they finish in seconds, so a multi-node workflow job can be seeded without depending on anything slow or absent. It also means the specs navigate under `/jobs/management/`, exercising a url segment other than the default one, which is where a missing entry in the type-to-segment map would hide.

##### The CI job

A job keyed `e2e` in the existing `.github/workflows/ci.yml`, reporting as the check **`ui-e2e`**. Lowercase and hyphenated to sit with `api-test`, `ui-lint` and `dev-env` rather than stand out, since the display name is what a required status check rule would be written against.

It runs on pull requests, like `dev-env`, and reuses the same `run_awx_devel` action with `build-ui: true` so it drives the built UI rather than a development server.

One step needs explaining. The action builds the UI *after* the web process is already serving, and Django caches the index template that names the bundle, so the job restarts `awx-uwsgi` and waits for the API to come back. Without it the tests run against whatever the image shipped rather than against the pull request. I found this the hard way: the suite reported the bug as unfixed against a UI that had the fix compiled into it.

The report and traces upload as an artifact on failure.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - UI

##### ASCENDER VERSION

```
25.5.1
```

##### ADDITIONAL INFORMATION

Verified both ways against a running environment:

```
against the UI without #760:   5 failed, 5 passed
  ✘ shows which of the workflow jobs is on screen
  ✘ switches to the job picked from the menu
  ✘ keeps working on a second move, without a page load in between
  ✘ lists every job node, and picking the current one is harmless
  ✘ filters the list by status

against the UI with #760:      10 passed (1.4m)
```

The five that fail are precisely the #742 regression specs, which is the demonstration that the suite earns its place.

It also caught something the jest suite and the development server both miss: changing the selector's label changed its lingui message id, and a **production** build renders an id with no catalog entry. The tab read `-t1kzb` instead of `Workflow Job 2/4` until the catalogs were extracted again. That fix is the second commit on #760.

Running it locally needs an environment that is already up:

```bash
make docker-compose
docker exec tools_awx_1 make ui-devel
cd awx/ui/e2e && npm ci && npx playwright install --with-deps chromium && npm test
```

`awx/ui/e2e/README.md` covers the configuration and how to add a spec.

It runs on pull requests, alongside `dev-env`, because that is where it catches things before they land.


blaipr:
Rebased. The conflict was entirely in `awx/ui/src/locales/*`, the generated catalogs, and it came from this branch sitting on top of #760: both regenerate them, and `main` had moved underneath.

This branch is stacked on #760 rather than on `main` directly, so it is rebased onto the rewritten #760 tip and its own diff is just the end to end suite:

```
 .github/workflows/ci.yml            |  69 +
 awx/ui/e2e/.gitignore               |   8 +
 awx/ui/e2e/README.md                |  59 +
 awx/ui/e2e/package-lock.json        |  78 +
 awx/ui/e2e/package.json             |  14 +
 awx/ui/e2e/playwright.config.js     |  31 +
 awx/ui/e2e/seed.js                  | 157 +
 awx/ui/e2e/tests/helpers.js         |  79 +
 awx/ui/e2e/tests/job-output.spec.js |  94 +
 awx/ui/e2e/tests/login.spec.js      |  22 +
 awx/ui/e2e/tests/tabs.spec.js       |  40 +
 awx/ui/eslint.config.mjs            |   3 +
```

`eslint .` and `prettier --check` are clean on it.

One thing worth noting given #760 grew a fix since: the toggle there is now a `styled(MenuToggle)`, and `workflowToggle` in `helpers.js` matches on `button[class*="menu-toggle"]`. styled-components merges its generated class with the PatternFly ones rather than replacing them, so the selector still matches and the specs did not need touching.

cigamit (approved):

PR #768, opened 2026-08-27
docs: Clean up references

ssimpson89:
##### SUMMARY

Fixes stale collection and plugin references in the docs: `awx.awx.tower` becomes `awx.awx.controller` where the docs describe generated inventory output, `ee_reference.rst` moves to `ctrliq.ascender`, and two `ansible/awx-ee` links now point at `ctrliq/ascender-ee`.

##### ISSUE TYPE

##### COMPONENT NAME
 - Docs

##### ASCENDER VERSION
```
25.5.2.dev108+gc3c2e51c0
```

### DIFF
--- a/docs/docsite/rst/userguide/inventory_plugins_templates.rst
+++ b/docs/docsite/rst/userguide/inventory_plugins_templates.rst
@@ -390,5 +390,5 @@ Red Hat Ansible Automation Platform
 
 	include_metadata: true
 	inventory_id: <inventory_id or url_quoted_named_url>
-	plugin: awx.awx.tower
+	plugin: awx.awx.controller
 	validate_certs: <true or false>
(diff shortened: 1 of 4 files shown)


docs/docsite/rst/userguide/inventory_plugins_templates.rst
```diff
@@ -390,5 +390,5 @@ Red Hat Ansible Automation Platform
 
 	include_metadata: true
 	inventory_id: <inventory_id or url_quoted_named_url>
-	plugin: awx.awx.tower
+	plugin: awx.awx.controller
```
cigamit, line 393: Should this not be ctrliq.ascender.controller?

ssimpson89: @cigamit I thought so too but Claude flagged that this is still what gets generated behind the scenes in [awx/main/models/inventory.py:1679-1686](https://github.com/ctrliq/ascender/blob/3799cb741f58a20c8cd24fe6673644cd29c0c5e6/awx/main/models/inventory.py#L1679) so I think its technically correct.

cigamit: Ah. I see.  Ya, that is something we will need to update.  I also need to update the vmware plugin to use the newer collection too, so I will do both at the same time.

cigamit (approved):

Issue #783, opened 2026-09-01
Azure AD authentication broken on 25.6


arjangit1234:
### Please confirm the following

- [x] I have checked the [current issues](https://github.com/ctrliq/ascender/issues) for duplicates.
- [x] I understand that Ascender is open source software provided for free and that I might not receive a timely response.

### Bug Summary

After upgrading to Ascender 25.6 the authentication through Azure AD is broken. When clicking the button to sign in with AzureAD the page gives a 405 error.

After downgrading tot 25.5.1 it works fine again.

### Ascender version

25.6

### Select the relevant components

- [x] UI
- [ ] UI (tech preview)
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

kubernetes

### Modifications

no

### Steps to reproduce

Enable authentication through AzureAD and click button to log on using Azure AD

### Expected results

Getting logged in in the Ascender UI.

### Actual results

Error 405 method not allowed

TheWitness:
Could you check the the logs and upload anything you find, obfuscate as required.  Thanks!!

brittyazel:
I had the same issue with OIDC auth, and CIQ support let me know that this issue shoudl be fixed by commit 15a585351b4f8a8f30d1dfe9c1dcc3453fc8f1f6 and that a new build should be coming out "very shortly". Just passing along the message.

galibozek:
I have the same problem with saml, can you please release the version since we have a fix?

cigamit:
A new version will be released today

galibozek:
Unfortunately I stopped getting the 405 but I still can't log in via saml (ee has the wrong tag in last release too)

arjangit1234:
I'm not able to install the latest release due to the ascender-ee being tagged as 26.5.1 instead of 25.6.1. Therefore i cannot test yet if this fixes the issue with AzureAD login.

  Warning  Failed     17s (x2 over 29s)  kubelet            Failed to pull image "ghcr.io/ctrliq/ascender-ee:25.6.1": rpc error: code = NotFound desc = failed to pull and unpack image "ghcr.io/ctrliq/ascender-ee:25.6.1": failed to resolve reference "ghcr.io/ctrliq/ascender-ee:25.6.1": ghcr.io/ctrliq/ascender-ee:25.6.1: not found
  Warning  Failed     17s (x2 over 29s)  kubelet            Error: ErrImagePull
[...]

PR #784, opened 2026-09-01
Start social auth logins with a POST form

samoz83:
##### SUMMARY
social-auth-app-django 6.0.1 (bumped in #731) decorates the login-initiation view with require_POST, but the login page still rendered each provider button as a plain link. Every SSO click got a 405 and no social provider could log in. Related #783.

The buttons now start the login by submitting a hidden form that carries `csrfmiddlewaretoken`, which is what social-app-django expects since its 5.5.0 login-CSRF hardening (an anonymous GET must no longer be able to start an SSO flow). The POST is a full page navigation rather than a fetch so the 302 out to the identity provider still works. `getCookie` already existed in `api/Base.js`, it is now exported and reused rather than duplicated.

##### ISSUE TYPE

##### COMPONENT NAME

 - UI

##### ASCENDER VERSION

```
25.6.0
```

##### ADDITIONAL INFORMATION

Reproduced on a 25.6.0 k8s install with Azure AD: clicking the SSO button
gives `HTTP 405, allow: POST`, and the web pod logs
`django.request Method Not Allowed (GET): /sso/login/azuread-oauth2/`.

Verified the POST path against the same live instance before patching:

```
$ curl -sc cj.txt https://<host>/api/login/ -o /dev/null
$ curl -sb cj.txt -X POST -H 'Referer: https://<host>/' --data-urlencode "csrfmiddlewaretoken=<token from cookie>" -o /dev/null -w '%{http_code} %{redirect_url}\n' https://<host>/sso/login/azuread-oauth2/ 
302 https://login.microsoftonline.com/common/oauth2/authorize?...&redirect_uri=https://<host>/sso/complete/azuread-oauth2/&...
```

The Login jest suite passes (18 tests), including a new test asserting the click submits a POST form containing the CSRF token.

### DIFF
--- a/awx/ui/src/api/Base.js
+++ b/awx/ui/src/api/Base.js
@@ -8,7 +8,7 @@ const updateStorage = debounce((key, val) => {
   window.dispatchEvent(new Event('storage'));
 }, 500);
 
-function getCookie(name) {
+export function getCookie(name) {
   const match = document.cookie.match(new RegExp(`(?:^|;\\s*)${name}=([^;]*)`));
   return match ? decodeURIComponent(match[1]) : null;
 }
--- a/awx/ui/src/screens/Login/Login.js
+++ b/awx/ui/src/screens/Login/Login.js
@@ -29,6 +29,7 @@ import {
 } from '@patternfly/react-icons';
 import useRequest, { useDismissableError } from 'hooks/useRequest';
 import { AuthAPI, RootAPI, MeAPI } from 'api';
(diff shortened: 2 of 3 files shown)


cigamit:
The one thing that I think is left out is the when users set the "Login redirect override URL" under "Miscellaneous Authentication settings", as that could still also require a POST (but its problematic, as it could also require a GET depending on the provider / authentication type).

But we can probably worry about that later and get this fix in now

cigamit (approved):

PR #786, opened 2026-09-02
Upstream 16620 - fix: check if the lock was acquired and return if not

blaipr:
Port of [ansible/awx@3917f3bf7](https://github.com/ansible/awx/commit/3917f3bf7) (ansible/awx#16620).

## The problem

`advisory_lock` in `awx/main/utils/pglock.py` wraps `django_pg_utils.locks.advisory_lock`, which runs `pg_try_advisory_lock` when `wait=False`, then yields the result and executes the body regardless:

```python
cursor.execute(sql, params)
acquired = cursor.fetchone()[0]
...
yield acquired
finally:
    if acquired: ...
```

`inspect_execution_and_hop_nodes` opened the lock without binding the yielded value, so it never skipped anything. `cluster_node_heartbeat` calls it on every control node, which means the full receptor status sweep and the `last_seen` writes ran on all of them concurrently, with no mutual exclusion.

Every other `wait=False` call site in the tree already gets this right: `json_migration_*` and `awx_periodic_scheduler` in the same module, `gather_analytics` in `awx/main/analytics/core.py`, and the task manager. This was the only one that did not.

## The change

Bind the yielded value and return early when the lock is held elsewhere, matching the surrounding call sites.

The timing and counter logging that upstream added in the same commit is deliberately left out, since it is diagnostics rather than the fix.

## Testing

Two unit tests in `awx/main/tests/unit/tasks/test_system.py`, adapted from upstream: the receptor controller is never reached when the lock is not acquired, and the sweep still runs when it is. Upstream passes `receptor_ctl` in as an argument, whereas this function calls `get_receptor_ctl()` itself, so the tests patch that instead.

`black --check awx` and `flake8 awx` pass.

### DIFF
--- a/awx/main/tasks/system.py
+++ b/awx/main/tasks/system.py
@@ -504,7 +504,11 @@ def inspect_established_receptor_connections(mesh_status):
 
 
 def inspect_execution_and_hop_nodes(instance_list):
-    with advisory_lock('inspect_execution_and_hop_nodes_lock', wait=False):
+    with advisory_lock('inspect_execution_and_hop_nodes_lock', wait=False) as acquired:
+        if not acquired:
+            logger.debug("Not running inspect_execution_and_hop_nodes, another instance holds lock")
+            return
+
         node_lookup = {inst.hostname: inst for inst in instance_list}
         ctl = get_receptor_ctl()
         mesh_status = ctl.simple_command('status')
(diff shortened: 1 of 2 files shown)


cigamit:
Not part of this PR, but just noting that Upstream has a few other changes in that file which we might look into, like wrapping that 
`mesh_status = ctl.simple_command('status')`
in a try/catch, etc...

cigamit (approved):

PR #793, opened 2026-09-02
Upstream 16466 - Fix TypeError when update_fields=None in model save methods

blaipr:
Port of [ansible/awx@f22df56e4](https://github.com/ansible/awx/commit/f22df56e4) (ansible/awx#16466).

## The problem

```python
update_fields = kwargs.get("update_fields", [])
```

The default only applies when the key is absent. `Model.save(self, ..., update_fields=None)` is the Django signature, so a caller that forwards `update_fields=None` explicitly gets `None` back from `.get`, and the next line does `update_fields.append(...)` or `if name not in update_fields`, which raises.

`awx/main/models/base.py` already reads `kwargs.get("update_fields") or []` in all three of its save methods, so this is the models deriving from it being inconsistent with their own base class, in ten places:

| File | Occurrences |
| ---- | ----------- |
| `awx/main/models/ad_hoc_commands.py` | 1 |
| `awx/main/models/inventory.py` | 1 |
| `awx/main/models/jobs.py` | 2 |
| `awx/main/models/notifications.py` | 1 |
| `awx/main/models/projects.py` | 1 |
| `awx/main/models/unified_jobs.py` | 2 |
| `awx/main/models/workflow.py` | 1 |
| `awx/conf/models.py` | 1 |

## Scope, stated plainly

This is hardening, not a fix for an observed failure. I went looking for a caller in this tree that passes `None` and did not find one: every `.save(update_fields=...)` site passes a list, and `_update_parent_instance_no_save` already guards with `if update_fields is None`. What is concretely true is that the base class and its subclasses disagree, and the subclasses are the fragile half.

`awx/conf/models.py` carries the same pattern and is included here. It is not part of the upstream commit.

`awx/dab/lib/abstract_models/common.py:115` also has it and is deliberately left alone: that tree is vendored from django-ansible-base, so it should be fixed upstream rather than diverge here.

## Testing

No behaviour changes for any existing caller, since `.get(k, [])` and `.get(k) or []` differ only when the value is present and falsy. `black --check awx` and `flake8 awx` pass.

### DIFF
--- a/awx/conf/models.py
+++ b/awx/conf/models.py
@@ -38,7 +38,7 @@ class Setting(CreatedModifiedModel):
         new_instance = not bool(self.pk)
         # If update_fields has been specified, add our field names to it,
         # if it hasn't been specified, then we're just doing a normal save.
-        update_fields = kwargs.get('update_fields', [])
+        update_fields = kwargs.get('update_fields') or []
         # When first saving to the database, don't store any encrypted field
         # value, but instead save it until after the instance is created.
         # Otherwise, store encrypted value to the database.
--- a/awx/main/models/ad_hoc_commands.py
+++ b/awx/main/models/ad_hoc_commands.py
@@ -214,7 +214,7 @@ class AdHocCommand(UnifiedJob, JobNotificationMixin):
         return AdHocCommand.objects.create(**data)
 
     def save(self, *args, **kwargs):
-        update_fields = kwargs.get('update_fields', [])
+        update_fields = kwargs.get('update_fields') or []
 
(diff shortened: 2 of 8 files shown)


cigamit:
Just a note, but `awx/main/models/inventory.py:494` has the same bug class untouched by both this PR and upstream: 

`'host_filter' in kwargs.get('update_fields', ['host_filter'])` raises `TypeError` if a smart-inventory save ever passes `update_fields=None` on postgres. Might be worth a follow up PR.

cigamit (approved):

PR #803, opened 2026-09-03
fix: give the job output scroll container a height independent of its rows

blaipr:
##### SUMMARY

Every job output page raised `ResizeObserver loop completed with undelivered notifications` on load, which the dev server's error overlay turned into a full-screen error.

The scroll container was `flex: 1 1 auto`, so its flex basis was its own content height, and it shared the column's free space with a footer strip that was `flex: 1` as well. react-virtual observes both the container and every row, and a row measurement changes the total height of the rows, so the first measurements resized the container in the same frame. Instrumenting the page on job 130 shows it: the container's rect went 347 to 268 to 346 px while its width stayed at 1022, one notice per load, and the same on the e2e fixture jobs. The row observations are delivered, the container's cannot be, and that is the notice.

The container now has a zero flex basis, so its height is the column's free space and nothing else, and it paints the gutter strip as its own background so the strip still runs to the bottom under short output. Two things go with that:

- the footer strip, since the gradient does its job and its share of the free space was half of the problem;
- the window error listener that tried to hide the notice from the overlay. A bubble-phase listener registered earlier still runs despite a later capture-phase `stopImmediatePropagation` (checked in Chromium), so it never worked, and there is nothing left to hide.

Nothing changes for a user of the page. The rendering is identical with output and in the empty-results state, in both themes, since the gradient uses the same variables the footer and wrapper used. The container no longer bounces while the first rows measure, the strip below a live job's first lines no longer moves as they arrive, and with the fix the container holds a single rect for the whole session.

##### ISSUE TYPE

##### COMPONENT NAME
 - UI

##### ASCENDER VERSION

```
25.6.1.dev21+g8408777933
```

##### ADDITIONAL INFORMATION

jsdom has no layout, so the cover is an end-to-end test: `watchLayoutLoops` installs an init script that records the notice before the app loads, and the new spec opens a fixture job's output, scrolls to the bottom and back, and asserts nothing was reported. Against the previous container it fails; with this change the whole job output spec passes against the dev server:

```
# previous container
  ✘  job output layout › measures its rows without a ResizeObserver loop
     Error: layout loops reported:
     ResizeObserver loop completed with undelivered notifications.

# this branch
  ✓  workflow job selector › shows which of the workflow jobs is on screen
  ✓  workflow job selector › switches to the job picked from the menu
  ✓  workflow job selector › keeps working on a second move, without a page load in between
  ✓  workflow job selector › lists every job node, and picking the current one is harmless
  ✓  workflow job selector › filters the list by status
  ✓  job output layout › measures its rows without a ResizeObserver loop
  6 passed
```

Under Node 22 with node_modules from the lockfile:

```
npm run test   Test Suites: 552 passed, 552 total / Tests:       2969 passed, 2969 total (251.856 s)
npm run lint   eslint . clean
prettier --check on JobOutput.js: clean; eslint on JobOutput.js: clean (the e2e directory is outside both tools' scope)
```

### DIFF
--- a/awx/ui/e2e/tests/helpers.js
+++ b/awx/ui/e2e/tests/helpers.js
@@ -57,6 +57,33 @@ function watchConsole(page) {
   };
 }
 
+// The job output measures its rows with a ResizeObserver, and a measurement
+// that resizes the scroll container in the same frame makes the browser report
+// "ResizeObserver loop completed with undelivered notifications" on window. It
+// is neither a console message nor a thrown exception, so watchConsole and
+// Playwright's pageerror both miss it; only an error listener installed before
+// the app loads sees it. Call this before the first navigation.
+async function watchLayoutLoops(page) {
+  await page.addInitScript(() => {
+    window.__ascenderLayoutLoops = [];
+    window.addEventListener(
+      'error',
+      (event) => {
+        if (/ResizeObserver loop/.test(event.message || '')) {
+          window.__ascenderLayoutLoops.push(event.message);
+        }
+      },
+      true
+    );
+  });
+  return {
+    async expectQuiet() {
+      const seen = await page.evaluate(() => window.__ascenderLayoutLoops);
+      expect(seen, `layout loops reported:\n${seen.join('\n')}`).toEqual([]);
+    },
+  };
+}
+
 // The tab bar hosts the workflow job selector, whose toggle shows either the
 // position within the workflow or, once a status filter is on, that status.
 const workflowToggle = (page) =>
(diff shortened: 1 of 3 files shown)


blaipr:
Fixed the duplicated gutter width raised in review.

The 85px value now lives in a single `GUTTER_WIDTH` constant, used both by the `ScrollContainer` background gradient and by the inline strip the virtualized content draws over its rows, so the two cannot drift apart.

Tested: `jest src/screens/Job/JobOutput` passes, 117 tests across 11 suites, and eslint is clean on the file.

cigamit (approved):

PR #806, opened 2026-09-03
feat: size the variables editors to their content, with a four-line floor

blaipr:
##### SUMMARY

The variables editors had a fixed height, six rows in the forms and four on the detail pages, with an expand button that opened the same editor in a 50vh modal. They now size to their content, never fewer than four lines and never more than fifty, and the expand button goes. Every form editor follows; on the detail side the template, job and workflow approval pages drop the row counts they passed, while inventory, host and credential type details keep theirs.

The height comes from ace's `minLines` and `maxLines` rather than a pixels-per-line formula, so the box matches the line height ace actually renders. The ceiling is there because ace renders every row it sizes itself to: `main` bounded that with `max-height: 90vh`, and dropping it along with the modal left 8,209 lines of host facts building 70,149 DOM nodes in a 197,026 px box.

## Ctrl+F

Ctrl+F in any of these editors threw and opened nothing, on `main` as well: ace asks for its search box through its own module loader, which resolves nothing under a bundler, and `ext-searchbox` was never imported. Importing it registers the extension up front. It belongs here because a capped editor scrolls, and ace's search reads the document rather than the rendered rows.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - UI

##### ASCENDER VERSION

```
25.6.2.dev4+g5a15cf1041
```

##### ADDITIONAL INFORMATION

The ceiling, measured on a host with 8,209 lines of facts, flipping only `maxLines` on the live page:

| | rendered lines | editor DOM nodes | height | vertical scrollbar |
| --- | --- | --- | --- | --- |
| capped at fifty | 52 | 434 | 1,210 px | yes |
| uncapped | 8,209 | 70,149 | 197,026 px | no |

An eight-line template renders eight lines with 5 px above the first and below the last, with the default font and with a serif override; a two-line one renders the four-row floor. Nine pages and four interactions, including the YAML toggle on those 8,209 lines, typing, and Ctrl+F, report no console errors or warnings.

Tests cover the fixed sizing, the auto sizing, the default ceiling and an explicit `maxRows`, and assert the expand button and its dialog are gone from both wrappers.

```
npm run test   552 suites, 2975 tests passed
npm run lint   eslint . clean
```


blaipr:
Fixed the dead wiring raised in review.

`ModeToggle` takes id, label, helpText, dataCy, mode, setMode and name, and never read `currentValue`, so passing it went nowhere. The prop is gone from the call site, and `currentValue` itself stays where it is still used, on the `CodeEditor` value below it.

Tested: `jest src/components/CodeEditor` passes, 19 tests across 3 suites, and eslint is clean on the file.

cigamit:
So there is a bit of an issue with this one.  For the Host Facts and Inventory Host Facts, it expands to fit the size of the facts, which in my case is over 8k lines.  This causes very noticeable slowness (~4 seconds to load on the screen) and a good bit of lag when scrolling.

blaipr:
Auto mode handed ace `maxLines: Infinity`, and ace renders every row it sizes itself to, so the editor built a line of DOM per line of facts. `main` never hit it because the editor carried `max-height: 90vh`, which clamped the box so ace only rendered a viewport's worth. That cap went out with the expand modal and nothing replaced it.

Fixed in 5a15cf1 with a fifty-row ceiling. On a host with 8,209 lines of facts, same page, flipping only `maxLines`:

| | rendered lines | editor DOM nodes | height |
| --- | --- | --- | --- |
| capped | 52 | 434 | 1,210 px |
| uncapped | 8,209 | 70,149 | 197,026 px |

It stays at 434 however long the facts are, and fifty rows is more than 90vh showed on a 1080p screen. Anything past row fifty scrolls inside the editor.

Separately in this branch: Ctrl+F in any variables editor was throwing and opening nothing, on `main` too, since `ext-searchbox` was never imported. It searches the whole document rather than the rendered rows, so it still finds a string on line 8,209.

Mind giving it another go on your host?

cigamit (approved):
Works much better now

PR #810, opened 2026-09-04
chore: remove the empty ui_next folder and the dead new-UI switch

blaipr:
##### SUMMARY

`awx/ui_next` was the mount point for the "next" UI, which Ascender never took. What was left of it was four files nothing used: a Makefile the root Makefile does not include, a README for a source tree that is not there, a placeholder index page, and a `urls.py` that `awx/urls.py` never included. The dev bootstrap script copied the placeholder into a build directory Django never served.

Around it sat the switch for that UI: a `UI_NEXT` setting, registered under System and defaulting to False, which migration 0186a already forces to false; and a `ui_next` flag in `/api/v2/config/` that only two banners read, on the dashboard and the subscription usage page, offering a link to `/ui_next`, a path that does not exist.

All of it goes: the folder, the bootstrap block, the `.gitignore` lines, the row in the copilot instructions, the setting and its default, the config flag, and both banners with the imports only they used. The message catalogs are left as they are; the banner's entry drops out of them the next time they are regenerated. The migration stays, as history; a `UI_NEXT` row left in `conf_setting` is inert once the setting is no longer registered.

Visible in the API: `/api/v2/settings/system/` no longer lists `UI_NEXT`, and `/api/v2/config/` no longer carries `ui_next`.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ASCENDER VERSION

```
25.6.2.dev1+g4262ff7980
```

##### ADDITIONAL INFORMATION

No reference to `ui_next` or `UI_NEXT` remains in the tree outside the migration. Checked inside the `ascender_devel` image and under Node 22:

```
awx-manage check                       System check identified 1 issue (1 silenced)   (unchanged)
awx-manage check_migrations --check    No changes detected
pytest awx/conf/tests, api/test_settings.py, api/test_root*.py   152 passed
jest src/screens/Dashboard src/screens/SubscriptionUsage        8 passed
npm run lint (eslint .)                clean
black, flake8 on the three Python files; bash -n on the bootstrap script   clean
```


cigamit:
Ya, the only reason I left it, was I was debating whether we wanted to split out the UI from python code.  But I found it creates too much of a hassle having to ensure you sync the proper versions when testing / creating.  I also see all the confusion it causes upstream, what a nightmare.

cigamit:
One thing to note, after this is merged, any old tree will need to manually delete the ui_next folder. It may contain some older artifacts, and since you removed the lines from .gitignore, those artifacts now can sneak into the next commit if they aren't watching.

cigamit (approved):

PR #811, opened 2026-09-04
build: move the container images to Python 3.14

blaipr:
##### SUMMARY

The container images build against Python 3.12. This moves them to 3.14.

Rocky 9's AppStream already carries it: `python3.14`, `python3.14-devel` and `python3.14-pip` come from AppStream, and `python3.14-setuptools` and `python3.14-packaging` from `crb`, which the Dockerfile enables already. No new repository is needed. Python 3.13 was the other candidate and was rejected: every one of those packages lives only in EPEL, which the production stages do not enable today.

`python3.12-psycopg2` is dropped rather than carried across. The venv is built with `include-system-site-packages = false`, so the RPM was never visible to the application, nothing on the system imports it, and the application uses `psycopg` 3 from `requirements.txt`. There is no `python3.14-psycopg2` and nothing needs one.

**Four things break on 3.14, and the interesting part is that three of them are silent at build time.**

1. **The multiprocessing start method.** `awx/main/dispatch/pool.py` used the default, which changed on Linux from `fork` to `forkserver` in 3.14. Workers inherit the loaded Django app registry from the parent, which only `fork` provides; under `forkserver` the child re-imports and raises `AppRegistryNotReady`. This affects the running service, not just tests. The context is now pinned to `fork`, matching what `awx/main/models/workflow.py:1286` already does. `tools/scripts/firehose.py` forks the same way after its `setup_django()` and is pinned alongside it.

2. **uwsgi's `py-call-osafterfork`.** Workers abort at fork with `Fatal Python error: PyMutex_Unlock: unlocking mutex that is not locked` and the master respawns them forever, so the web tier never comes up. The option calls `PyOS_AfterFork_Child` in a state 3.13+ rejects ([unbit/uwsgi#2710](https://github.com/unbit/uwsgi/issues/2710)), and uwsgi's pending fix ([#2736](https://github.com/unbit/uwsgi/pull/2736), a draft for over a year) skips the call on 3.13+ anyway, so the option is dead weight there. Found by @cigamit testing this branch. It is the one failure the suite could not have caught, since pytest never starts uwsgi.

3. **`zope-interface` 7.0.3 cannot be imported on 3.14 at all.** Defining `IElement` in its own `interfaces.py` raises `InvalidInterface: Concrete attribute, __classdictcell__`, from the class dict change in 3.13. It reaches us through `twisted`, so this would break the running service too. It compiles from source without complaint, which is exactly why a wheel-availability audit does not catch it.

4. **`test_postprocess_auth_basic_enabled`** seeded `__file__` via `locals().update()`. Since PEP 667 in 3.13, `locals()` inside a function returns an independent snapshot, so the update never reached the frame and `include()` raised `KeyError: '__file__'`. It now passes an explicit scope dict, which behaves identically on every version.

The requirements move is the minimum 3.14 forces, and the chain is worth recording. `pydantic-core` 2.16.2 builds through `pyo3` 0.20.2, which predates 3.14. cp314 wheels start at `pydantic-core` 2.35.0, which needs `pydantic` 2.12 or newer, which requires `typing-extensions>=4.14.1`. The `typing-extensions==4.12.2` pin was the real ceiling, so all three had to move together. `typing-inspection` arrives as a new dependency of pydantic and brings its license file with it.

```
pydantic          2.6.1  -> 2.13.5
pydantic-core     2.16.2 -> 2.46.5
typing-extensions 4.12.2 -> 4.16.0
typing-inspection        -> 0.4.4  (new)
zope-interface    7.0.3  -> 8.6
```

Regenerated with `requirements/updater.sh` under 3.14. No other pin moved.

##### ISSUE TYPE

##### COMPONENT NAME

- API
- Other

##### ASCENDER VERSION

```
awx: 25.6.2.dev0+g4deeb17b95.d20260904
```

##### ADDITIONAL INFORMATION

Verified in an image built from this checkout with `make docker-compose-build`:

- Full suite on 3.14: **3968 passed, 10 skipped**. The same suite on `main` under 3.14 fails exactly two, `TestWorkerPool::test_worker_processing` and `test_postprocess_auth_basic_enabled`, which are the two this branch fixes. Collection is identical at 3978 either way, so nothing was skipped into passing.
- `make test-postgres` against a real PostgreSQL: 4 passed.
- `awx-manage check_migrations --dry-run --check`: no changes detected.
- `black --check awx`: 929 files unchanged.
- `flake8 awx`: clean.
- `yamllint -s .`: clean.
- Import smoke test across 46 top-level dependencies on 3.14: 46/46.
- `awx/` carries no imports of any module removed in 3.13 or 3.14: no PEP 594 modules, no `pkgutil.find_loader`, no `typing.ByteString`.

The suite covers the dispatcher pool, as the `main` baseline above shows, but it never starts uwsgi. The service paths were checked against a running 3.14 image as well:

- uwsgi serving `awx.wsgi` off the real `/etc/tower/uwsgi.ini`: all 5 workers spawn and reach `WSGI app 0 ready`. With `py-call-osafterfork` restored, the same command reproduces @cigamit's abort loop, so the option is the cause rather than a coincidence.
- A `fork` context child of `awx.main.dispatch.pool` inherits the app registry (`apps.ready` true, exit 0), where the default `forkserver` context kills the child.
- `twisted` 26.4.0 imports with `zope-interface` 8.6, including `zope.interface.interfaces.IElement`.

Motivation for going to 3.14 now rather than later: `django-guid` 3.6.1 caps `django<6.0` for `python_version >= "3.10" and python_version < "3.13"`, while allowing `django<7.0` on 3.13 and newer. That cap blocks any move to Django 6 on 3.12.

Rebased onto current `main` after #807, #808 and #809 merged, so the branch is two commits on top of it with no merge commits. `requirements.txt` was regenerated with `requirements/updater.sh` under 3.14 rather than merged textually, since a text merge would have reinstated `django-crum`, `django-extensions` and `djangorestframework-yaml`, which those PRs removed; the regenerated file is byte-identical to the rebase result.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -20,7 +20,7 @@ Supporting containers also running:
 - `tools_valkey_1` – Valkey (Redis-compatible) cache
 
 Runtime versions (inside container):
-- **Python** 3.12.12
+- **Python** 3.14.7
 - **Django** 5.2.14
 - **psycopg** 3.1.18
 - **black** 26.5.1 | **flake8** 7.3.0 | **yamllint** 1.38.0
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 # Modifications Copyright (c) 2023 Ctrl IQ, Inc.
 
-PYTHON := $(notdir $(shell for i in python3.12 python3; do command -v $$i; done|sed 1q))
+PYTHON := $(notdir $(shell for i in python3.14 python3; do command -v $$i; done|sed 1q))
 SHELL := bash
 DOCKER_COMPOSE ?= docker compose
 OFFICIAL ?= no
--- a/README.md
+++ b/README.md
@@ -2,14 +2,14 @@
 
 [![CI](https://github.com/ctrliq/ascender/actions/workflows/ci.yml/badge.svg)](https://github.com/ctrliq/ascender/actions/workflows/ci.yml)
 [![License](https://img.shields.io/badge/license-Apache--2.0-blue.svg)](./LICENSE)
-[![Python](https://img.shields.io/badge/python-3.12-blue.svg)](https://www.python.org/)
+[![Python](https://img.shields.io/badge/python-3.14-blue.svg)](https://www.python.org/)
 
 Ascender provides a web-based user interface, REST API, and task engine built on top of [Ansible](https://github.com/ansible/ansible). It is the automation platform at the centre of this project, based on the upstream [AWX](https://github.com/ansible/awx) project and maintained by Ctrl IQ.
 
 ## Requirements
 
 - Docker with Compose, for the development environment
-- Python 3.12, matching what the container images build against
+- Python 3.14, matching what the container images build against
 - A Kubernetes cluster for production, which the installer can provision for you
 
 ## Installation
--- a/awx/main/dispatch/pool.py
+++ b/awx/main/dispatch/pool.py
@@ -9,8 +9,7 @@ from datetime import datetime, timezone
 from uuid import uuid4
 
 import collections
-from multiprocessing import Process
-from multiprocessing import Queue as MPQueue
+from multiprocessing import get_context
 from queue import Full as QueueFull, Empty as QueueEmpty
 
 from django.conf import settings
@@ -29,6 +28,12 @@ if 'run_callback_receiver' in sys.argv:
     logger = logging.getLogger('awx.main.commands.run_callback_receiver')
 else:
     logger = logging.getLogger('awx.main.dispatch')
+
+# Workers inherit the loaded Django app registry from the parent process, which only
+# the fork start method provides. Python 3.14 changed the Linux default to forkserver,
+# where the child re-imports and raises AppRegistryNotReady, so the context is pinned
+# here rather than left to the platform default.
+_mp_ctx = get_context('fork')
 
 
 class NoOpResultQueue(object):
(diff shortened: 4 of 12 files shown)


cigamit:
Tested this.  While the container would build, the application wouldn't start.  This error was thrown over and over.
```
tools_awx_1       | DAMN ! worker 1 (pid: 487) died, killed by signal 6 :( trying respawn ...
tools_awx_1       | Fatal Python error: PyMutex_Unlock: unlocking mutex that is not locked
```

The simple fix was this.
```
diff --git a/tools/ansible/roles/dockerfile/files/uwsgi.ini b/tools/ansible/roles/dockerfile/files/uwsgi.ini
index 0f27eb9b..8aed627f 100644
--- a/tools/ansible/roles/dockerfile/files/uwsgi.ini
+++ b/tools/ansible/roles/dockerfile/files/uwsgi.ini
@@ -13,7 +13,10 @@ buffer-size = 32768
 harakiri = 115
 harakiri-graceful-timeout = 110
 harakiri-graceful-signal = 6
-py-call-osafterfork = true
+; py-call-osafterfork is fatal on Python 3.13+ with uwsgi <= 2.0.31: workers die
+; at fork with "PyMutex_Unlock: unlocking mutex that is not locked"
+; (https://github.com/unbit/uwsgi/issues/2710). uwsgi's pending fix skips the
+; PyOS_AfterFork_Child call on 3.13+ anyway, so the option is dead there.
 
 if-env = UWSGI_MOUNT_PATH
 mount = %(_)=awx.wsgi:application
 ```
 
 Seems they still a PR in uwsgi for this, but its still a draft and hasn't been merged in over a year.
 https://github.com/unbit/uwsgi/pull/2736

blaipr:
Applied, thank you. Confirmed it here first: same image and same ini with that line the only difference, and the abort loop appears and disappears with it. Your patch is in as-is in 449e440.

It also pointed me at a second instance I had missed. `tools/scripts/firehose.py` forks after `setup_django()` with the default context, which is the same bug this branch fixes in `awx/main/dispatch/pool.py`, so 3.14 kills it the moment it spawns. Pinned to `fork` in c592c7f.

Rebased onto current `main`, and the full compose stack is now verified end to end on a rebuilt image: 14/14 supervisor programs up, 5 uwsgi workers steady with no respawns, dispatcher and callback receiver forking their children, `/api/v2/ping/` serving on 3.14.7. Suite is 3968 passed and 10 skipped, where `main` under 3.14 fails exactly the two this branch fixes.

cigamit (approved):

PR #813, opened 2026-09-04
build: run the test suite on PostgreSQL

blaipr:
##### SUMMARY

The suite ran on SQLite so that no service was needed. That hid whole code paths: the partitioned event tables, the `COPY` that assembles job stdout, and every query-count assertion behave differently or not at all there, and Ascender is deployed against PostgreSQL. `settings_for_test_pg` existed to opt a single file into a real database; the suite now uses one throughout.

**Four defects came out of the move, two of them in code that ships.**

**1. `assertNumQueries` has been inert against our PostgreSQL backend.** `awx/main/db/profiled_pg/base.py` reimplements `force_debug_cursor` as a property so it can consult the `awx-manage profile_sql` flag, and gave it a setter that throws the value away:

```python
@force_debug_cursor.setter
def force_debug_cursor(self, v):
    return
```

Django's `CaptureQueriesContext`, which `assertNumQueries` and `django_assert_max_num_queries` are built on, turns recording on by assigning to exactly that attribute. Every query-count assertion therefore observed **zero** queries and could only pass by expecting zero. Invisible while the suite ran on SQLite's stock backend. Once fixed, one budget in `test_bulk.py` turned out to be genuinely marginal and is documented there.

**2. `inventory_import` took a lock in order to reject its arguments.** The `--source` check sat inside `advisory_lock(...)`. On SQLite that helper yields without touching the database; on PostgreSQL it takes a real advisory lock. Moved up beside the other option validation.

**3. `is_testing()` silently disabled a `post_migrate` hook.** It compared `DJANGO_SETTINGS_MODULE` for equality and looked for `'py.test'` in `argv[0]`. Under any other test settings module, and in xdist workers whose argv is not the pytest entry point, it returned `False`, disabling the hook that creates the `ServiceID`. It surfaced as `Expected ServiceID to be created in data migrations`, which points at migrations rather than the cause. Now a prefix match, and `pytest` is accepted.

**4. The stdout `COPY` was never exercised.** The `sqlite_copy` fixture's `MockCopy` ignored the SQL entirely and returned every event's stdout, so none of this was tested:

```python
where_parts.insert(0, sql.SQL('job_created = {}').format(sql.Literal(self.created)))
```

the partition filter, `COPY`'s row separator, or its `{a,b}` array rendering. Those fixtures are deleted and the tests run the real query.

The remaining changes are SQLite assumptions: primary keys that only start at 1 because SQLite resets `rowid` on rollback, SQLite's exact `UNIQUE constraint failed:` wording, and a live connection count the SQLite branch hardcoded to `1`.

##### Speed

```
SQLite, before                                  347s
PostgreSQL, naive                               890s
request-scoped transaction.atomic()             603s
cheap password hasher in tests                  379s
after removing the dead sqlite_copy fixtures    320s
```

**The first is the one that matters.** `conftest` calls the view directly, bypassing Django's handler, so production's `ATOMIC_REQUESTS: True` never applied. A request whose statement failed left the whole test's transaction aborted and broke every later request in that test, which production never does. Wrapping the call restores production's shape and removed the need for 61 tests to opt out of transactional isolation.

**The hasher is the other.** One PBKDF2 hash costs ~0.7s and the fixtures create a user for nearly every functional test. Hashing is not what any of them test.

Tried and **rejected on measurement**: `--dist=load` is 19% slower than `loadfile` because per-test distribution loses fixture reuse; `-n 16` is within noise of `-n auto` and adds a flake.

##### ISSUE TYPE

##### COMPONENT NAME

- API
- Other

##### ASCENDER VERSION

```
awx: 25.6.2.dev0+g4deeb17b95.d20260904
```

##### ADDITIONAL INFORMATION

Verified on **Python 3.12 with Django 5.2.17**, neither changed here: **3950 passed, 6 skipped**. Skips drop from 10 to 6 because four tests guarded on `connection.vendor` for PostgreSQL-only SQL now actually run. The one failure in the full run, `test_refresh_token_expiration_is_respected`, is a pre-existing flake: it passes on its own, three times out of three. `black --check awx`, `flake8 awx` and `yamllint` are clean.

The connection comes from the settings the development container already loads out of `/etc/tower/conf.d`; anywhere without those files, such as a throwaway CI container, each part can be supplied through `AWX_TEST_DATABASE_*`.

**Two things a reviewer should weigh.**

The CI service wiring could not be exercised locally, since GitHub Actions cannot run here. It is yamllint-clean and conventional, but the `services: postgres` block plus `--network host` is reasoned rather than observed. The `AWX_TEST_DATABASE_*` path it depends on **is** verified: every timing above was produced by pointing the suite at a separate PostgreSQL container through exactly those variables.

`test_project_list_ordering_with_duplicate_names` now asserts only that repeated requests agree. With every name identical there is no tie-break in the `ORDER BY`, so PostgreSQL may return the rows in any order and asserting one would be asserting undefined behaviour. That has a product implication worth a separate look: a paginated list ordered by a non-unique column has no stable order, so pages can in principle repeat or skip rows. A `pk` tie-break in list ordering would fix it, but that is an API behaviour change and does not belong here.


cigamit:
### Fresh dev environments can't create the test database

The suite now needs to create `test_awx_pg` (and the `test_awx_pg_gwN` copies under xdist), but in the standard docker-compose dev environment the `awx` role has neither `SUPERUSER` nor `CREATEDB`:

```
$ docker exec tools_postgres_1 psql -U postgres -tc "SELECT rolname, rolsuper, rolcreatedb FROM pg_roles WHERE rolname IN ('awx','postgres');"
 postgres | t | t
 awx      | f | f
```

So an out-of-the-box `make test` inside `tools_awx_1` fails immediately:

```
Got an error creating the test database: permission denied to create database
```

The only `ALTER ROLE awx CREATEDB;` in the tree was in the dev-env CI step this PR deletes, and nothing in `tools/docker-compose` provisioning grants it. CI is unaffected — the new `services: postgres` block uses `POSTGRES_USER: awx`, which makes the role a superuser of that instance — and any machine where the grant was already applied by hand won't notice, which is presumably why this didn't surface during development.

Suggested fix: grant it during docker-compose provisioning so the dev environment works out of the box, e.g. as part of the postgres bootstrap:

```sql
ALTER ROLE awx CREATEDB;
```

(or at minimum document the one-liner in the testing docs:
`docker exec tools_postgres_1 psql -U postgres -c 'ALTER ROLE awx CREATEDB;'`)

For what it's worth, after applying that grant the full suite here is green: **3972 passed, 6 skipped in 2:33** on Python 3.14.7 / Django 5.2.17, with `check_migrations` and the linters clean.

blaipr:
Fixed in 3a68e8a. The grant now comes from a start hook that `run-postgresql` sources on every container start, so an existing volume picks it up on its next restart rather than having to be recreated, and `ALTER ROLE` is idempotent.

Verified by first revoking the grant here, since I had applied it by hand exactly as you describe and had therefore been testing on a machine that could not see the bug. After recreating the container: `=> sourcing /opt/app-root/src/postgresql-start/grant-createdb.sh`, `rolcreatedb = t`, and `CREATE DATABASE` as the `awx` role succeeds.

Also addressed the review in 2501129f5f: the hasher comment quoted Django 6.1's iteration count, left over from this branch's first life as a Django bump, and `ordered_nodes` stayed a queryset that `values_list()` never evaluates, so each index lookup issued its own query.

Still open, and thanks for the nudge to look: `api-migrations` fails on PostgreSQL. `apply_initial_migration(('main', None))` leaves `main_jobevent_host_id_b03b6059` behind, so replaying forward to 0180 hits `relation already exists`. The partitioned event table migrations do not reverse symmetrically and SQLite never exercised that path. Working on it.

cigamit:
Also, while you are working on the `api-migrations`

### Minor leftovers, none blocking

Follow-up material rather than review blockers — fine to land as-is or fold into this PR, whichever you prefer.

**1. Stale references to the SQLite suite and the deleted `make test-postgres` target**

- `awx/main/tests/functional/commands/test_cleanup_jobs_postgres.py:9` — the docstring still says "Run these with `make test-postgres`", but this PR deletes that target. These tests now just run as part of `make test` (the vendor skipif is always true).
- `awx/main/tests/functional/commands/test_callback_receiver.py:128` — "tests use sqlite3 which will not raise an error" no longer describes the suite.
- `awx/main/tests/functional/test_migrations.py:12` — the docstring says the smoke test ensures migrations work "when sqlite is the backing database (vs. the default DB of postgres)"; the migrator now runs against PostgreSQL like everything else.

**2. Dead SQLite branch in shipping code**

`awx/api/serializers.py:2371`:

```python
if settings.DATABASES and ('sqlite3' not in settings.DATABASES.get('default', {}).get('ENGINE')):
    # sqlite acts different with bulk_create -- it doesn't return the id of the objects
```

The comment says outright it existed to appease the test suite. Now that no supported configuration runs SQLite, the condition is always true and the guard (plus its comment) can go, making `url` unconditional.

**3. Remaining `'credential_type': 1` literals in test_credential.py**

Four are left (the `*_xfail` tests around lines 102, 116, 168, 182). They're harmless — each asserts a 403, and permission is denied before the credential type is ever resolved — but since the PR replaced every other hardcoded pk for exactly this class of assumption, replacing these too would finish the job and keep the file consistent.

blaipr:
`api-migrations` is fixed in 3055629, and the follow-ups are in c7bb8b4.

One root cause behind two failures, both from the five `_unpartitioned_*` tables that 0144 creates by renaming with raw SQL. They back proxy models, so Django introspection does not list them, and `django_test_migrations` builds its clean state by dropping only the tables Django knows about. The orphans survive it, and then:

- a renamed table keeps its original index names, so the replay dies on `relation "main_jobevent_host_id_b03b6059" already exists`
- they carry real foreign keys into `main_host`, `main_job` and the other job tables, and PostgreSQL will not `TRUNCATE` a table something outside the list still references, so the `transactional_db` flush fails at teardown with `cannot truncate a table referenced in a foreign key constraint`

Dropping them on both sides of the replay clears both, and the fixture depends on `transactional_db` so its teardown is ordered ahead of that flush. `1 failed, 3 errors` to `2 passed`, confirmed here against PostgreSQL on 3.14.

Worth recording that I only found the second one because fixing the first changed the result to `2 passed, 2 errors` rather than green. I had been piping these runs through `tail`, which was discarding the traceback.

All three of your follow-ups are in, including the `test_migrations.py` docstring, which I had already caught while working on the fixture. The serializer guard is gone: as its own comment admitted it only ever existed for the SQLite suite, so it was already dead in production. The four `credential_type` literals now use `credentialtype_ssh.id` like the rest of the file. 114 passed across the four affected files.

cigamit (approved):

PR #814, opened 2026-09-05
build: move to Django 6.1.1

blaipr:
##### SUMMARY

Django 6.1 is the current release. The `<6.0` cap was ours: a floor with a ceiling attached, kept because `django-guid` 3.6.1 pins `django<6.0` for Python 3.10 to 3.12, which stops applying once the images build against 3.14 (#811). Django 6.1 also requires SQLite 3.37 against Rocky 9's 3.34.1, which is why the suite had to move to PostgreSQL first (#813).

```
django              5.2.17 -> 6.1.1
djangorestframework 3.17.2 -> 3.18.0
```

**DRF has to move with it.** 3.17.2 does

```python
from django.utils.cache import cc_delim_re
```

which 6.1 removed **without a release note**, so `awx/urls.py` simply fails to import. 3.18.0 declares support for 6.1.

##### Two deprecations, both fixed rather than deferred

Neither is loud, which is why they are handled here rather than left for later.

**`from_db()` gained a `fetch_mode` keyword**, which Django records on the instance state:

```python
@classmethod
def from_db(cls, db, field_names, values, *, fetch_mode=None):
    ...
    if fetch_mode is not None:
        new._state.fetch_mode = fetch_mode
```

An override that drops it opts the model out of fetch modes entirely, so `CredentialType` and everything inheriting the vendored dab override would have silently ignored `FETCH_PEERS` and `FETCH_RAISE`. Both overrides now forward it. The dab one carries an `# Ascender:` marker for future re-syncs.

**`BaseEmailBackend.fail_silently` is deprecated.** Passing it up to `super().__init__()` is itself the deprecated act, and Django's own message says a subclass that supports it must set its own attribute. All eight notification backends now do.

##### And one that turned out not to be Django's fault

Chasing the last remaining warning found a test issuing `select_related()` with no arguments, which means "follow every non-null foreign key". The cause was the `mock_access` fixture: a `MagicMock` attribute is truthy *and* iterates empty, so `optimize_queryset`'s guard passed while supplying no fields.

```python
if access_class.select_related:                                    # MagicMock -> truthy
    queryset = queryset.select_related(*access_class.select_related)  # unpacks to nothing
```

The fixture has been doing that on Django 5.2 as well; 6.1's deprecation only made it audible. It now carries the empty tuples a real access class has.

Between them, these three take the suite from **2113 warnings to 0**.

##### ISSUE TYPE

##### COMPONENT NAME

- API
- Other

##### ASCENDER VERSION

```
awx: 25.6.2.dev0+g6e10fb75b3.d20260905
```

##### ADDITIONAL INFORMATION

Verified in an image built from this checkout, against PostgreSQL: **3950 passed, 6 skipped, 0 failed, 0 warnings**. `black --check awx` (928 files), `flake8 awx` and `yamllint` clean.

Rebased onto `main` after #807, #808 and #809 merged. `requirements.txt` was regenerated with `requirements/updater.sh` under 3.14 rather than merged textually, since a text merge would have reinstated `django-crum`, `django-extensions` and `djangorestframework-yaml` which those PRs removed. The regenerated file is byte-identical to the rebase result.

**Worth weighing before merging:** Django 6.1 leaves extended support in **December 2027**, four months *earlier* than 5.2 LTS's April 2028. So this is not a move that buys a longer support horizon. The case for taking it now is that it absorbs the 6.0 and 6.1 breaking changes while nothing is urgent, which makes the hop to 6.2 LTS in April 2027 a small one, and that fetch modes are aimed squarely at the serializer N+1 behaviour, which the `from_db` fix above is a prerequisite for using.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -16,7 +16,7 @@ docker exec tools_awx_1 bash -c "cd /awx_devel && <command>"
 ```
 
 Supporting containers also running:
-- `tools_postgres_1` – PostgreSQL database (used by the running app; tests use SQLite)
+- `tools_postgres_1` – PostgreSQL database (used by the running app and by the test suite)
 - `tools_valkey_1` – Valkey (Redis-compatible) cache
 
 Runtime versions (inside container):
@@ -57,7 +57,7 @@ docker exec tools_awx_1 bash -c "cd /awx_devel && yamllint -s ."
 
 ## Running Tests
 
-Tests use SQLite (not PostgreSQL) and in-memory channel layer — no database setup needed. Settings: `awx.main.tests.settings_for_test` (configured in `pytest.ini`). Default pytest flags: `--reuse-db --nomigrations --tb=native --timeout=300`.
+Tests run against PostgreSQL, the same backend Ascender deploys on, with an in-memory channel layer. Settings: `awx.main.tests.settings_for_test` (configured in `pytest.ini`). Default pytest flags: `--reuse-db --nomigrations --tb=native --timeout=300`.
 
 ### Unit tests (~15 seconds, run frequently)
 ```bash
@@ -119,7 +119,7 @@ Migration files live in `awx/main/migrations/` (218 existing files).
 | `awx/main/migrations/` | 218 Django migrations |
 | `awx/main/tests/unit/` | Fast isolated unit tests (~1139 tests) |
 | `awx/main/tests/functional/` | Django TestClient-based API tests (~1987 tests) |
-| `awx/main/tests/settings_for_test.py` | Test settings (SQLite, in-memory cache) |
+| `awx/main/tests/settings_for_test.py` | Test settings (PostgreSQL, in-memory cache) |
 | `awx/settings/` | Settings modules: `defaults.py`, `development.py`, `production.py` |
 | `awx/sso/` | SSO/LDAP/SAML backends + tests |
 | `awx/ui/` | UI / React frontend (npm) |
--- a/.gitignore
+++ b/.gitignore
@@ -14,8 +14,6 @@ pre-commit-user
 # Tower
 awx-dev
 awx/settings/local_*.py*
-awx/*.sqlite3
-awx/*.sqlite3_*
 awx/job_status
 awx/projects
 awx/job_output
@@ -93,7 +91,6 @@ htmlcov
 pep8.txt
 scratch
 testem.log
-awx/awx_test.sqlite3-journal
 .pytest_cache/
 
 # Mac OS X
--- a/Makefile
+++ b/Makefile
@@ -103,7 +103,6 @@ clean-api:
 	rm -rf .tox
 	find . -type f -regex ".*\.py[co]$$" -delete
 	find . -depth -type d -name "__pycache__" -delete
-	rm -f awx/awx_test.sqlite3*
 	rm -rf requirements/vendor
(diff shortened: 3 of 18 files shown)


blaipr:
Conflicts resolved and both review comments addressed.

The conflict was ancestry, not content: the branch carried its own copies of the Python 3.14 move and the PostgreSQL suite change, both of which had already landed on `main` as squash commits. Cherry-picked this branch's two real commits onto current `main` and dropped the duplicates, 30+ files down to 13.

`api-test` failed after the suite passed, in `awx-manage check_migrations`: it runs on the default settings, and Django 6.1 needs SQLite 3.37 where Rocky 9 ships 3.34.1. The SQLite fallback is gone rather than worked around, so the default is PostgreSQL.

Review: the `djangorestframework` floor was 3.17.2 while the lock moved to 3.18.0, and 3.18.0 is the first release declaring Django 6.1 support. `from_db` is a classmethod that took `self`, now `cls`.

`_sqlite_helper.py` and its migrations, `split_ids_for_sqlite`, and the now dead engine guards stay for a cleanup of their own.

cigamit:
### What the PR missed

- The email backend in `/awx/main/notifications/email_backend.py#L27` subclasses Django's SMTP backend, and Django 6.1 deprecates constructing that directly. Every email notification send now emits a `RemovedInDjango70Warning ("Use mail.mailers instead")`. No test constructs it, so the "0 warnings" figure is true of the suite but not of the shipping code. It breaks in Django 7.0, so it needs a MAILERS-based rework before then. Note that this warning is invisible from `awx-manage shell -c`, since Django treats frames inside its own module namespace as internal and suppresses them.

### Nits for the author

- The fallback introduces a third database env-var naming scheme (`AWX_DATABASE_*`, alongside `DATABASE_*` in the docker-compose environment template and `AWX_TEST_DATABASE_*` in test settings).

- `Makefile#L106` and the `.gitignore` entries for `awx_test.sqlite3` are now dead. The copilot instructions still say tests use SQLite, though that predates this PR.

blaipr:
All three in 682e40d.

The email backend is the real one, thank you. Confirmed against the 6.1.1 wheel: the warning fires on `"alias" not in kwargs`, and `warn_about_external_use` resolves an effective caller inside Django, which is exactly why `awx-manage shell -c` shows nothing. Your point about the "0 warnings" figure being true of the suite and not the shipping code is the part worth keeping: nothing constructs one in a test, so the number was never evidence.

I have documented it at the class rather than silenced it. There is no MAILERS rework available here: `MAILERS` is a static dict keyed by alias, while a notification template carries its own host, port and credentials decrypted per send in `NotificationTemplate.send`, so there is no alias to point at, and `create_connection` offers no runtime registration. `get_connection` is deprecated on the same timer, so there is no third door. Passing `alias=None` would suppress the warning, since the check is only the kwarg's presence, but it hides a warning that is correct about 7.0 and would break anyway once the deprecated branch goes. This needs solving before 7.0 and it is a change of its own.

`AWX_DATABASE_*` is gone, plain values instead. You are right that it was a third scheme, and worse, a dead one: nothing read it, and CI only worked because the literals happen to match its service.

The `.gitignore` patterns and the Makefile clean rule are out, and the copilot instructions now say PostgreSQL. Removing the ignore patterns surfaces stale `awx.sqlite3` files in older checkouts, which seems like the right outcome.

cigamit (approved):

PR #816, opened 2026-09-05
Keep max_retries when a workflow node is created without a template

fernandorocagonzalez:
### Problem

`POST /api/v2/workflow_job_template_nodes/` silently drops `max_retries` when the node is created without a `unified_job_template`. No error, no warning: it answers 201 with the default value.

| Call | Sent | Stored |
| --- | --- | --- |
| POST without ujt | `max_retries=7` | `max_retries=0` |
| POST with ujt | `max_retries=7` | `max_retries=7` |
| PATCH afterwards | `max_retries=7` | `max_retries=7` |

`identifier` and `all_parents_must_converge` travel in that same POST and survive it, and they are node fields exactly like `max_retries` is.

It only bites a client that creates the node first and attaches the template in a second call. The UI never does that for regular nodes, it always sends the template in the same POST.

### Cause

In `LaunchConfigurationBaseSerializer.validate()`, when there is no template to validate the prompted fields against, everything is thrown away except a hardcoded list of three field names. Dropping `limit` there is right, it is a prompt and cannot be validated without a template. `max_retries` is a plain node field. #564 added it to `Meta.fields` of both node serializers and missed this list.

### Fix

Add `max_retries` to that list, plus a comment saying what the branch is for so the next plain field added to the node has a better chance of landing in the right place.

### Tests

Three new ones in `test_workflow_node.py`, next to the ones from #564:

- POST without a template with `max_retries` gets stored
- the same POST with `limit` still drops the prompt, that behaviour is correct and stays
- POST without a template, then PATCH the template in, and the value is still there at the end

Found while testing the ascender collection against a 25.6.0 controller.

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -4288,8 +4288,11 @@ class LaunchConfigurationBaseSerializer(BaseSerializer):
         elif self.instance:
             ujt = self.instance.unified_job_template
         if ujt is None:
+            # Without a template we cannot validate any of the prompted fields, so only
+            # the node's own fields are kept. Anything added to the node itself belongs
+            # on this list, prompts do not.
             ret = {}
-            for fd in ('workflow_job_template', 'identifier', 'all_parents_must_converge'):
+            for fd in ('workflow_job_template', 'identifier', 'all_parents_must_converge', 'max_retries'):
                 if fd in attrs:
                     ret[fd] = attrs[fd]
             return ret
--- a/awx/main/tests/functional/api/test_workflow_node.py
+++ b/awx/main/tests/functional/api/test_workflow_node.py
@@ -98,6 +98,43 @@ def test_node_max_retries_rejects_out_of_range_values(inventory, project, workfl
 
 
 @pytest.mark.django_db
+def test_node_max_retries_settable_without_template(workflow_job_template, post, admin_user):
+    # max_retries belongs to the node, not to the template, so a POST that does not
+    # name a template yet has no reason to drop it
+    url = reverse('api:workflow_job_template_workflow_nodes_list', kwargs={'pk': workflow_job_template.pk})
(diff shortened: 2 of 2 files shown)


cigamit:
I might be missing something, but why would you set max_retries on an approval node?  You certainly can't via the UI, so not sure why you would via the API either.

fernandorocagonzalez:
> I might be missing something, but why would you set max_retries on an approval node?  You certainly can't via the UI, so not sure why you would via the API either.

Not an approval node, a workflow node, I think your have crossed with the other prs? 

What happens, and that's also something that we could discuss is, if you create a workflow object, but, you don't setup the flow content, then the option is lost, and it should not, it should be unrelated since you can do that via the GUI

Or maybe you think the approval node is going to inherit?

fernandorocagonzalez:
Wait actually, Im re reading that and Im making some tests , I think its me the one confused

fernandorocagonzalez:
Well it was me the confused ok (I think I cross messed up something with the ansible collection for ascender )

Bad example on my part. max_retries on an approval node does nothing anyway, retry_pending() bails out for approvals (workflow.py:512) and the UI hides the field there. Let forget about approval nodes

The real bug is just that a POST without unified_job_template returns 201 and silently drops max_retries, while all_parents_must_converge and identifier on the same request survive. Same kind of node field, just missed in #564. Only shows up if you create the node first and attach the template later.

fernandorocagonzalez:
You're right, bad example on my part. max_retries on an approval node does nothing anyway, `retry_pending()` bails out for approvals (`workflow.py:512`) and the UI hides the field there.

The real bug is just that a POST without `unified_job_template` returns 201 and silently drops `max_retries`, while `identifier` and `all_parents_must_converge` on the same request survive. Same kind of node field, just missed in #564. Only shows up if you create the node first and attach the template later.

Description rewritten and the approval test swapped for a create then attach one.

cigamit (approved):

PR #859, opened 2026-09-07
build: bump incremental from 24.7.2 to 24.11.0 in /requirements

blaipr:
Bumps [incremental](https://github.com/twisted/incremental) from 24.7.2 to 24.11.0.

Transitive pin, reached through `twisted`, which uses it for its own version handling.
Nothing in `awx/` imports it.

```
-incremental==24.7.2
+incremental==24.11.0
```

## Checks

- Full `TEST_DIRS` run: **3986 passed, 6 skipped**.
- `pip check` reports no broken requirements.


cigamit (dismissed):

blaipr:
Conflict is resolved, this is mergeable again.

I regenerated the lockfile against current main with `updater.sh upgrade incremental` rather than resolving the conflict by hand, because the correct diff is not just the version line: incremental 24.11.0 swaps its `setuptools` dependency for `packaging`, so two `# via` entries move with it. Hand merging would have left generated output whose comments no longer match the dependency graph.

Full `TEST_DIRS` run is re-running on the current base, and I will flag it here if anything turns up.

cigamit (approved):

blaipr:
Re-run on the current base is green: 3986 passed, 6 skipped.

PR #897, opened 2026-09-08
fix: move the OAuth2 refresh and ID token models into the main app

blaipr:
> ## ⚠️ Alternative to #895 — this one removes the cause, not the message
>
> #895 adds `oauth2_provider.W011` to `SILENCED_SYSTEM_CHECKS`, and its reading of
> the situation is correct: nothing is broken, the cross-app dependencies are
> declared by hand, and the warning is log noise. It is one line and it works.
>
> **This PR instead moves the refresh and ID token models into the `main` app, so
> the check has nothing to report.** After it, `awx-manage check` is clean with no
> entry in `SILENCED_SYSTEM_CHECKS`.
>
> They are mutually exclusive: pick one. If #895 merges first, this PR should drop
> the `oauth2_provider.W011` entry it added. The honest trade is below.

## The trade against #895

| | #895 | this PR |
| --- | --- | --- |
| Change | one settings line | two models, two settings, 8 call sites, 1 migration |
| Risk | none | touches the token tables on upgrade |
| Result | warning hidden, split remains | split gone, warning cannot recur |
| If DOT adds more checks on this shape | silence each one | already correct |

If the appetite is for the smallest safe change, #895 is the better PR and I would
say so. This one is worth it only if the split itself is worth removing.

## What the warning is about

django-oauth-toolkit points `AccessToken.source_refresh_token` at the refresh token
model and `RefreshToken.access_token` back at the access token model. AWX swapped
only the access token into `main`, leaving that pair a circular foreign key across
two apps.

Both models now live in `main`, so the cycle is contained:

```
OAUTH2_PROVIDER_REFRESH_TOKEN_MODEL   main.OAuth2RefreshToken
OAUTH2_PROVIDER_ID_TOKEN_MODEL        main.OAuth2IDToken
```

Neither adds a field over the upstream abstract base. They exist so the access
token's own foreign keys resolve inside one app. `Grant` stays where it is: it
points at the application one way and is not part of the cycle.

## Why the models are declared in 0025 and not in a new migration

This is the part worth reviewing.

Swapping a model rewrites how **every historical** migration state resolves
`OAuth2AccessToken.source_refresh_token` and `.id_token`. `main/0041` adds the
first of those fields, `main/0183` the second, and `oauth2_provider/0004` adds both
to its own (swapped out) `AccessToken`. All three need the models to already exist,
or the state cannot be built at all:

```
ValueError: The field main.OAuth2AccessToken.source_refresh_token was declared with
a lazy reference to 'main.oauth2refreshtoken', but app 'main' doesn't provide model
'oauth2refreshtoken'.
```

Creating them in a new migration and forcing the order with `run_before` fixes fresh
installs and breaks every existing one, because `run_before` is a dependency and the
migrations it must precede are already applied:

```
InconsistentMigrationHistory: Migration oauth2_provider.0001_initial is applied
before its dependency main.0214_oauth2_token_models_into_main
```

Declaring them in `0025` avoids all of it. That migration already creates the
application and access token the two models point at, already carries
`run_before ('oauth2_provider', '0001_initial')`, and is applied on every existing
install, so it introduces no new ordering and cannot make an applied migration
depend on an unapplied one.

`0214` therefore changes no state at all. It only reconciles the database an older
install actually has with the models `0025` describes, and does nothing on a fresh
install.

## Three things 0214 has to handle

All three come from the same root: once the model is swapped out, Django skips
operations on it, so DOT's later migrations never touch the old table.

- **`token_checksum` is missing** from the old table. DOT adds it in its own 0015,
  which is skipped. It is NOT NULL on the new model.
- **`resource` is missing** for the same reason, from DOT's 0018.
- **pgcrypto is not guaranteed to be installed**, so `digest()` is unavailable and
  the sha256 is computed in Python, matching what DOT's own backfill produces.

Rather than chase DOT's columns one release at a time, the copy fills any required
column the old table lacks from the field's own default, and computes
`token_checksum` properly since a default would collide on the
`(token_checksum, revoked)` unique constraint.

## Verified on both paths

**Fresh install**, empty database:

```
all 295 migrations apply
tables created: main_oauth2idtoken, main_oauth2refreshtoken
oauth2_provider_refreshtoken / _idtoken: never created
source_refresh_token_id -> main_oauth2refreshtoken
id_token_id             -> main_oauth2idtoken
awx-manage check: no issues
```

**Upgrade**, against a copy of a real database carrying 71 access tokens and a live
refresh token:

```
access tokens preserved: 71
refresh tokens moved:     1
token_checksum computed:  029003e7... == sha256("probe-refresh-token")
old tables:               dropped
source_refresh_token_id -> main_oauth2refreshtoken
id_token_id             -> main_oauth2idtoken
awx-manage check: no issues
```

## Checks

- Full `TEST_DIRS` run: **3986 passed, 6 skipped**, including the three migration
  tests that replay history to 0180, 0188 and 0189.
- `black --check` and `flake8` clean.
- `awx-manage makemigrations --check --dry-run`: no changes detected.


blaipr:
Both review points fixed.

**0025 imports.** Confirmed the concern is real: after `import oauth2_provider`, neither `oauth2_provider.models` nor `oauth2_provider.generators` resolves, so the file was relying on something else importing them first. Both are now imported explicitly, matching 0198 and 0213, which also fixes the same latent problem for the pre-existing `generators` references.

**0214 checksum.** The unique key is `(token_checksum, revoked)` and `revoke_family()` stamps a whole family with one timestamp, so a shared placeholder can collide. The constraint does reject it:

```
duplicate key value violates unique constraint
Key (token_checksum, revoked)=(, 2020-01-01 00:00:00+00) already exists
```

The checksum is now computed in the INSERT with core PostgreSQL `sha256()`, available since 11 and needing no pgcrypto, so there is no window where two rows hold the same placeholder. It agrees with `hashlib.sha256` and with the toolkit's own backfill.

Re-verified on a copy of a real database seeded with a revoked family of three tokens sharing one timestamp: migrate OK, each row carrying its own correct checksum.

In fairness to the review: I could not get the previous code to fail in practice, only the constraint violation directly. Computing it in the INSERT is the safer form either way, so I took it rather than leaving the window open.

cigamit (approved):

PR #898, opened 2026-09-08
feat: allow an administrator to upload a UI theme

blaipr:
Addresses the forum request in
[choose your own UI theme](https://forum.ascender-automation.org/d/18-choose-your-own-ui-theme/2):
a way to add a theme without rebuilding the image.

## The problem

The themes in `awx/ui/src/themes` are discovered by `require.context` and bundled
at build time, so nothing an administrator does can add to them. Getting a new
one in front of a user currently means rebuilding and redeploying.

## What this adds

Two settings in the UI category, alongside `CUSTOM_LOGO` and `CUSTOM_TITLE`, so
the settings API and the Settings screen pick them up with no further plumbing:

| setting | holds |
| ------- | ----- |
| `CUSTOM_THEME` | the contents of a `.css` file |
| `CUSTOM_THEME_NAME` | what to call it in the theme list, `Custom` if blank |

The stylesheet reaches the browser in a `<style>` element when the config loads,
so it costs no extra request, and it is appended to the head so it comes after
the bundled themes and wins at equal specificity. It then appears in the theme
list beside the shipped ones and is selected and persisted by exactly the same
code, including the per-user `preferred_theme`.

A theme scopes itself the way the shipped ones do:

```css
/* Name: Solarized */
html[data-theme="custom"] { --pf-t--global--background--color--primary--default: #fdf6e3; }
```

and for a dark theme:

```css
html.pf-v6-theme-dark[data-theme="custom"] { ... }
```

Darkness is detected with the same rule `config/themeMetaLoader.js` applies to
the shipped themes, so an uploaded theme and a built in one are judged alike.

## What is rejected, and why

Only what reaches the network. The value is written exclusively by a system
administrator and only ever placed in a `<style>` element, so it cannot introduce
script; the concern is different:

- **`@import`** fetches a second stylesheet from wherever it points. That turns a
  theme into a request to a third party on every page load, and moves the real
  content somewhere this validation never sees.
- **A remote `url()`** (`https:`, `http:` or protocol relative `//`) does the same
  for fonts and images, and leaks the viewer's address and headers to that host.

Relative paths and `data:` URIs stay allowed, which is what the shipped themes
already use for their own fonts. There is also a 5 MB cap, against roughly 60 KB
for the largest shipped theme, so it is a guard against a runaway paste rather
than a real limit.

## One structural choice worth a look

`setCustomTheme` lives in a new `src/customTheme.js` rather than inside
`themeRegistry.js`. The registry cannot be imported under Jest, because it uses
webpack's `require.context`, which is why `testUtils/themeRegistryMock.js` exists
as a hand written copy of its behaviour.

Nothing about the custom theme needs the bundler, so keeping it separate makes it
directly testable, and the mock now calls the real implementation instead of
duplicating it, so that part can no longer drift.

## Checks

- 13 backend tests over the field: what it accepts, what it rejects and why, the
  size cap, and the name.
- 6 UI tests over installing, replacing and removing the stylesheet, the default
  name, and dark detection.
- Full `TEST_DIRS` run: **3999 passed, 6 skipped**.
- **135 UI tests across the 29 suites** that consume the config context pass.
- `black`, `flake8`, `prettier` and `eslint` clean.

### DIFF
--- a/awx/main/tests/functional/test_custom_theme.py
+++ b/awx/main/tests/functional/test_custom_theme.py
@@ -1,0 +1,67 @@
+import pytest
+
+from awx.conf import settings_registry
+
+
+@pytest.fixture
+def custom_theme_field():
+    return settings_registry.get_setting_field('CUSTOM_THEME')
+
+
+@pytest.mark.parametrize(
+    'css',
+    [
+        pytest.param('html[data-theme="custom"] { --pf-t--color: #fff; }', id='plain_css'),
+        pytest.param('@font-face { src: url(../../public/static/fonts/Inter.woff2) format("woff2"); }', id='relative_url'),
+        pytest.param('a { background: url(data:image/png;base64,iVBORw0KGgo=); }', id='data_uri'),
+        pytest.param('', id='empty_clears_the_theme'),
+    ],
+)
+def test_accepts_a_self_contained_stylesheet(custom_theme_field, css):
+    """Anything that stays inside the deployment is allowed through untouched."""
+    assert custom_theme_field.to_internal_value(css) == css
+
+
+@pytest.mark.parametrize(
+    'css, expected',
+    [
+        pytest.param('@import url("https://example.invalid/theme.css");', 'may not use @import', id='import_https'),
+        pytest.param('@IMPORT "other.css";', 'may not use @import', id='import_is_case_insensitive'),
+        pytest.param('a { background: url(https://example.invalid/x.png); }', 'may not reference remote URLs', id='https_url'),
+        pytest.param('a { background: url(http://example.invalid/x.png); }', 'may not reference remote URLs', id='http_url'),
+        pytest.param('a { background: url(//example.invalid/x.png); }', 'may not reference remote URLs', id='protocol_relative_url'),
+        pytest.param("a { background: url('https://example.invalid/x.png'); }", 'may not reference remote URLs', id='quoted_remote_url'),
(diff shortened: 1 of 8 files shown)


blaipr:
Both review points fixed, and both were real.

**CSS comment bypass.** Confirmed against the old code: `url(/*x*/https://evil.invalid/a.png)` was accepted. A parser treats a comment as whitespace inside `url()`, so it reaches that host exactly as the plain form does. Comments are now stripped before the `@import` and remote URL checks; the value is still stored exactly as written, only the checks see the stripped copy. Four regression tests cover it.

One thing that looks like a bypass and is not, recorded so nobody adds a check for it: `@/*x*/import` is not an import at all, since an at-keyword is `@` followed immediately by an identifier, so a comment there breaks the token. A comment in front of the at-rule, `/*c*/@import`, is real and is caught.

**Saved custom theme never restored.** Also real, and the cause is as described: `App.js` applies the stored theme before these settings exist, so a saved custom theme is unknown at that point, `applyTheme` falls back to the default and writes `default` into sessionStorage on the way past. Reading the session value back here returned that fallback, so the theme was installed and then never selected. It now reads the saved preference from localStorage through `getSavedThemeId`, which that early call does not touch because it does not persist.

Full TEST_DIRS run: 4003 passed, 6 skipped. UI tests 6 passed, prettier and eslint clean.

cigamit (approved):

PR #910, opened 2026-09-08
build: replace black and flake8 with ruff

blaipr:
Rebased onto #909 now that it has merged, so this is three commits and no longer carries a copy of it.

The rest of [Path to Ruff](https://forum.ascender-automation.org/d/44-path-to-ruff), steps two and three taken together because neither is much use without the other. black goes, flake8 goes, and `[testenv:linters]` is left holding ruff and yamllint.

## The reformat

`ruff format` changes 106 of the 931 Python files under `awx`, 381 lines added and 369 removed. It is the same 931 files black was checking, so this is a like for like reformat rather than a widening of what gets touched. It is the first commit on its own so the tooling swap that follows is readable.

Most of the diff is the `u''` string prefix, which ruff drops where black's `skip-string-normalization` left it alone, and which has meant nothing since Python 3.0. The rest is ruff joining implicitly concatenated strings that fit on one line, dropping redundant parentheses, and using the parenthesized form for multiple context managers.

Checked rather than assumed: every one of the 106 files parses to the same AST before and after, once `Constant.kind` is normalised out, that being the only field the `u` prefix sets. Nothing in the reformat changes behaviour.

## target-version turns out to be a formatter setting

Worth knowing before anyone raises it: ruff reads `target-version` for the formatter, not only for the lint rules. At `py314` it rewrites 31 files from `except (A, B):` to the PEP 758 `except A, B:`, which is a syntax error on every older interpreter. `py39` through `py313` produce identical output, so the default is left in place and `[tool.ruff]` now carries a comment saying why. It matters here because ruff infers the target from `requires-python`, and `[project]` only arrived in #900, so the field that would trigger it is one line away.

## What follows upstream and what does not

AWX has already made this move, so the shape is theirs: a `make format` target driven by `RUFF_FORMAT_ARGS`, `black` kept as an alias so the habit and any script that calls it still work, a `lint` target beside it, and a pre-commit hook that honours both `AWX_IGNORE_BLACK` and the new `AWX_IGNORE_RUFF`.

Two things deliberately do not follow them. Ruff stays pinned at 0.16.6 in `[testenv:linters]` like every other linter here, where upstream leaves it floating. And the rule selection stays at the 48 ported in #909 rather than their 12.

## Smaller things in the same change

- `[tool.black]` is removed. Its two settings moved to `[tool.ruff]` in #909, and `fast = true` has no equivalent because ruff format does not run black's AST equivalence check in the first place.
- tox no longer calls `make`, so `allowlist_externals` and the `BLACK_ARGS` and `PYTHON` setenv entries go with it.
- The Copilot instructions listed 15 flake8 rules where the block selected 48, so that file has been understating the linting since well before this change. Corrected while the section was being rewritten anyway.

## Verification

- `tox -e linters` clean, which is the command the api-lint CI job runs.
- `make check`, `make format`, `make lint`, `make black` and `make api-lint` all work.
- Full suite over `awx/main/tests/unit`, `awx/main/tests/functional`, `awx/conf/tests` and `awx/sso/tests`: **4003 passed, 6 skipped**.


awx/main/models/credential/__init__.py
```diff
@@ -1036,7 +1036,7 @@ def create(self):
                 'label': gettext_noop('Username'),
                 'type': 'string',
                 'help_text': gettext_noop(
-                    'CIQ Ascender Automation Platform username id to authenticate as.' 'This should not be set if an OAuth token is being used.'
+                    'CIQ Ascender Automation Platform username id to authenticate as.This should not be set if an OAuth token is being used.'
```
cigamit: Missing space between sentences

awx/main/models/credential/__init__.py
```diff
@@ -1050,7 +1050,7 @@ def create(self):
                 'label': gettext_noop('OAuth Token'),
                 'type': 'string',
                 'secret': True,
-                'help_text': gettext_noop('An OAuth token to use to authenticate with.' 'This should not be set if username/password are being used.'),
+                'help_text': gettext_noop('An OAuth token to use to authenticate with.This should not be set if username/password are being used.'),
```
cigamit: Missing space between sentences

awx/main/models/credential/__init__.py
blaipr: Good catch, fixed in f2bcbf4416. It predates this PR: the two halves were adjacent string literals, so the value always ran the sentences together, and joining them into one literal is what made it visible. Kept as its own commit since it changes a string value, which the reformat commit does not. The catalogue entries for both are already fuzzy from the Red Hat to CIQ rebrand, so no translation is lost.

awx/main/models/credential/__init__.py
blaipr: Good catch, fixed in f2bcbf4416. It predates this PR: the two halves were adjacent string literals, so the value always ran the sentences together, and joining them into one literal is what made it visible. Kept as its own commit since it changes a string value, which the reformat commit does not. The catalogue entries for both are already fuzzy from the Red Hat to CIQ rebrand, so no translation is lost.

blaipr:
@cigamit thanks, both fixed in f2bcbf4416.

They predate this PR. The two halves were adjacent string literals, so the concatenated value always ran the sentences together. Merging them into one literal is what made it visible, and main has both halves sitting adjacent on the same line already.

Kept as its own commit, since it changes a string value and the reformat commit deliberately changes none.

Nothing is lost in translation either. The catalogue entries for both are already marked fuzzy from the Red Hat to CIQ rebrand, so they are not used at runtime, and the existing Spanish and Japanese renderings break the two sentences properly, which is the other way of saying the space belongs there.

cigamit (approved):

PR #506, opened 2026-06-25
Upgrade PatternFly from v4 to v5

blaipr:
## SUMMARY

Upgrade PatternFly from v4 to v5. This is the first step in the PatternFly modernization roadmap (PF4 → PF5 → PF6).

**What changed:**

- Bump all `@patternfly/*` packages from v4 to v5
- Migrate Dropdown/KebabToggle/DropdownToggle to PF5 MenuToggle/DropdownList API (24 files)
- Migrate deprecated PageHeader to PF5 Masthead/MastheadToggle/MastheadMain/MastheadContent/PageToggleButton
- Update CSS class prefixes: 
    - `pf-c-*` → `pf-v5-c-*`
    - `--pf-global--` → `--pf-v5-global--`
    - `pf-theme-dark` → `pf-v5-theme-dark`
    - `pf-screen-reader` → `pf-v5-screen-reader`
- Fix PF5 Tooltip async state updates causing `act()` warnings with a `settleTooltips()` test helper
- ESLint: enable `allowAsProps` for `react/no-unstable-nested-components` (PF5 toggle render prop pattern)

**Not included (will open separate PRs):**

~20 files still import from `@patternfly/react-core/deprecated`_

- Select (~15 files)
- Wizard (3 files)
 
These work (PF5 ships them as a transitional API) but the composable replacements have significantly different APIs and need careful per-component migration.

## ISSUE TYPE

- New or Enhanced Feature

## COMPONENT NAME

- UI

## ADDITIONAL INFORMATION

- 552 test suites pass, lint clean
- Headless browser smoke test: 30/30 checks pass (login, 20 screens navigated, PF5 CSS classes verified, toolbar/sidebar render, dark mode toggle, zero API errors)
- 251 files changed, +2708 / −2292

PR #919, opened 2026-09-11
build: convert the UI to TypeScript

blaipr:
Converts `awx/ui` to TypeScript in full: every file under `src` and `testUtils`, the 552 test files included. No JavaScript is left in either tree, `allowJs` is off, and `tsc --noEmit` is clean under `strict` with `noUncheckedIndexedAccess`.

Stacked on #916, which moves the UI onto Vite. This branch starts at that pull request's head, so it should land after it.

## How the types are set up

- `tsconfig.json` checks but never emits: Vite strips the types, so this is a checker rather than a build step.
- `strict` and `noUncheckedIndexedAccess` are on from the first commit. Nothing was converted under a looser setting and tightened later.
- `src/types/api.generated.ts` comes from the platform's own OpenAPI schema, via `npm run generate-api-types`. Responses are described from the schema rather than by hand.
- Each component exports a props interface named after it. A prop is optional where the component defaults, guards or forwards it, and required where the component calls or dereferences it itself.
- Nothing under `src` or `testUtils` is `any`. Where a shape cannot be described it is `unknown`, narrowed where it is read. `@typescript-eslint/no-explicit-any` is on, and the one place the rule is suppressed says why: TypeScript only recognises a class as a mixin base when its constructor takes `any[]`.
- A partial fixture in a test is cast at the point of use. The application's own types were never loosened to accept one.

## What the checker found

Fourteen behaviours were wrong and are fixed here. Every one predates this branch, and each is described where it is fixed.

- The metrics line chart passed d3 the event where the datum was expected, and reached for `d3.select(this)` inside arrow functions, where `this` is not the element.
- `WorkflowJobTemplate.helptext` was keyed by API field names while the form, the webhook subform and the detail screen all read camelCase, so no workflow help popover had ever rendered.
- The workflow node type pickers passed `role_level` as `read()`'s second argument, which it ignores, so the job, workflow and system job lists offered templates the user cannot execute.
- Job output dropped the socket `connectJobSocket` returns and called `closeWebSocket()` with nothing, so the connection outlived the screen.
- `TeamRolesList` read the organization admin count off the response rather than off its `data`, so `isAdminOfOrg` was always false. Its test asserted the same wrong shape.
- Both execution environment screens and the user token form assigned the help text factory instead of calling it.
- Four lookups called formik's `setTouched()` with no argument, which is a no-op, and the team form passed it a field name.
- `CredentialTypeDetails` wrapped a caught error in `new Error(error)`, which loses the response the dialog reads.
- A weekly schedule's weekday could not be unticked: seven checkboxes read PatternFly 6's `(event, checked)` as PatternFly 5's `(checked, event)`.
- The disassociate permission message never appeared, because the predicate chosen for the check was always truthy and the other branch always ran.
- Three sites compared against `('a' || 'b')`, which is `'a'`: an ad hoc shell command with no arguments was accepted, one field's checks never ran, and the notification Test button stayed enabled while a test was pending.
- Eight topology tooltips spelled their Cypress hook `dataCy`, which PatternFly spreads onto the DOM as `datacy`, so no `[data-cy=...]` selector matched.
- `NodeModal` gave Formik a prop named `onSave`, which Formik has no such prop for, so the form had no submit handler.
- `PlaybookSelect` asked PatternFly 6 for a `"default"` toggle status, which it dropped, and `CredentialPluginField` used `styles.modifiers.error`, which it also dropped.

Dead code came out with them: an unreachable branch calling a `ProjectsAPI.readCredentials` that does not exist, a `sourceOptions` prop eleven inventory subforms never read, and a dozen other props passed to components that do not declare them.

Typing the tests turned up another six, all of them tests that were not testing what they said:

- `new Error({ response })` hands the object to the Error constructor, which stringifies it, so the errors those tests passed carried no response and the response branch never ran.
- `LaunchPrompt` called `CredentialTypesAPI.loadAllTypes` instead of stubbing it, against an already mocked module, so the credential types were never there.
- Two render helpers returned `{ history, ...utils }`, overwriting the history the test had just created.
- A schedule test spread the schedule over the inventory override it meant to apply.
- A visualizer action still carried the i18n object the reducer stopped taking when it moved to a start label.
- `ConstructedInventoryHint` replaced the whole config module, leaving the render helper without the provider it reads at import time.

## Naming what was left

The conversion left 3,259 values in the sources and 1,009 in the tests carrying a named `any`, so that what had not been described yet was greppable rather than invisible. The last commits take all of them, sources and tests alike, and each wave found something:

- Nine list screens passed PatternFly no row index, so every row's checkbox was labelled "Select all rows" and the selection cell was not announced as a cell at all. Two tests asserted that as the expected shape.
- The launch prompt's labels field set the caught error as the field's error message and rendered it, so a failed label lookup threw "Objects are not valid as a React child" rather than showing anything.
- `AnsibleSelect` wrote its name onto the change event whether it had one or not, so three selects that do not use it set the DOM attribute to the string "undefined".
- Reading a summary field the API had not sent threw rather than rendering nothing, in twenty-four places.
- Two role lists ran `parseInt` over an id the API already sends as a number.
- Four forms held their content error in a state initialised to `false` and set it to the error object.
- `getSearchableKeys` answers with one shape that four different declarations guessed at: `unknown[]`, `string[]`, `string[][]` and an untyped one. It is a named `SearchableKey` now, and the related keys beside it are plain field names.
- The jobs were three shapes behind one marker: a list row is a `UnifiedJob`, a job's own screen reads a `Job`, and what the sparkline draws are the summary field's `RecentJob` entries.

`summary_fields` gains the members the UI reads and the hand-written types did not name, and three fields the serializers build in method fields the schema cannot see: an organization's `galaxy_credentials`, a user's `auth`, and the `webhook_key` the project form fetches from its own endpoint.

Narrowing the rest of the sources found another nine, in the same way and all of them older than this branch:

- A plugin backed credential input was handed whether its value was valid and then derived that flag from itself, which is a string either way, so a value the form had already rejected still rendered as valid.
- The node type step took the launch configuration where it reads the node being edited, which it has never been given, so the alias a node already carries was never required and could be blanked.
- The peers list copied only the hostname and the node type off each peer's instance, so the four details the row opens to were always empty.
- An instance group's associate handler filtered for a node that was not control or not hop, which is every node. The list the modal offers already leaves both kinds out, which is why nothing went wrong.
- A boolean filter's toggle read a chip's `label`, which no chip has: the text is its `node`, so the toggle never showed what the filter was set to.
- Two settings screens wrote each setting's value onto the OPTIONS block itself, which every settings screen shares through context, so one screen's values reached the next screen's fields.
- `getScheduleUrl` and the launched by detail took a job of one of two kinds where either screen may be showing any of the six.
- The launch and relaunch endpoints answer a launch configuration, which the mixins described as two loose objects, so the button reached through an untyped promise for the job it had just started.
- Three schema fields are serializer method fields the generator can only describe as strings: an instance group's capacities and its job and instance counts, and an instance's `health_check_pending`, which the lists poll on.

The last of the sources and then the tests turned up eleven more:

- `DraggableSelectedList` took an `onRowDrag` prop, read it nowhere and rendered no drag handles. The drag-and-drop went out with the PatternFly v4 to v5 upgrade (#506) and was never restored. It works again, in the last commit below.
- The ad hoc command's review step reads `values.credential` where the wizard's field is `values.credentials`, and the plural key is excluded from the generic detail list above it, so the credential the user picked appeared nowhere on the step that exists to show what will run. Inherited from AWX 22.5.0.
- `JobTemplateForm` declared `handleSubmit` as the caller's value-taking submit, but `withFormik` injects its own over it. It was cast away twice to compile.
- `WorkflowOutputNode` read `prior_run_succeeded` and `prior_run_elapsed` off an index signature, because the node type is built from the workflow job *template* node, which carries neither.
- A node's elapsed time was formatted without converting it: the api sends the decimal as a string, and `secondsToHHMMSS` takes a number. Two fixtures held it as a number, which is what had hidden it.
- `module_name`'s OPTIONS choices are `[string | number | null, string]` pairs, not `[string, string]`, so the ad hoc wizard took the api's own choice type and six inventory lists stopped widening it.
- The project add and edit submits rewrote their values in place to turn each lookup's object into an id. They build the request body beside the values now. Their tests handed the mock form raw numeric credential ids, a shape the form cannot produce.
- Mock forms that stand in for a real one now take that form's own props type. Doing that to `JobTemplateEdit`'s stub showed its submit values clearing the execution environment to `''` where the form clears it to `null`, and carrying the api's nulls for three fields the form holds as empty strings.
- `WorkflowJobTemplateEdit`'s submitted values were missing `webhook_key`, one of the form's own fields.
- The organization form's instance groups stub called `onChange` with a second argument naming the field; the real lookup passes one.
- Three survey fixtures built a question with no `variable`, which is not a question the api can return.

The host add and edit screens parked their stubbed form's submit values on the global as an untyped slot, and the edit screen's failing-submit case overwrote it with a host payload where form values belong. The dates test carried an i18n stub nothing used, kept alive by an eslint-disable. Each mocked endpoint's response is named after the endpoint itself now, rather than `ApiResponse<any>`.

## Two things put right rather than recorded

**Reordering a selected list works again.** PatternFly 6 has no replacement for the `DataList` drag props the old implementation was built on: it keeps `DataListControl` and `DataListDragButton` as the pieces and leaves the dragging to the page. The rows carry native HTML5 drag now, the way `SurveyReorderModal` in this tree already does under PatternFly 6, and the four lookups that were handing `sortSelectedItems` to a prop nobody read are wired back up.

Two faults in the pre-#506 implementation are not carried over. Its drag handle set both `aria-label` and `aria-labelledby`, and labelledby wins, so every handle was announced as the row's text rather than as what it does; each is now "Reorder <name>". And it passed the keyboard instructions as `aria-describedby` text, where that attribute takes an id: they are rendered once per list and every handle points at them. The keyboard reorder those instructions promise is implemented rather than promised: space or enter picks a row up, the arrow keys move it, escape puts it back.

**The dead source types are gone.** The inventory source form carried subforms for `insights` and `rhv`, and the project form one for `insights`. The platform comments all three out of `SOURCE_CHOICES` and `SCM_TYPE_CHOICES`, so the API refuses every one: the subforms could not be reached, and the five list filters offering those types could only ever return nothing. Both subforms and the filters go, with two documentation URLs and three test fixtures that still listed them.

The `insights` credential kind is untouched, because the platform still has that one, and so is the job detail's scm type label, which names what a row holds rather than offering a choice.

## The message catalogues

Four of the guards added while typing sat inside a `Plural`'s `value`, and lingui names a placeholder after the expression it is handed: a bare identifier becomes `{name}`, anything else becomes `{0}`. So `sessionCountdown ?? 0` turned `{sessionCountdown}` into `{0}`, `failedSources` turned `{0}` into `{failedSources}`, and `forks ?? 0` did it twice. A changed message id matches nothing in the catalogues, so each one silently dropped that string back to English in all eight translated locales. None of the four guards was reachable, so all four are gone and the ids are as they were.

`npm run check-strings` now fails when the catalogues drift from the source, and `ui-lint` runs it, which makes it part of a check that already has to pass. It extracts against a throwaway catalogue so the working tree is untouched either way, and compares ids rather than files, because the catalogues carry source references with line numbers and comparing them whole would fail on any change that shifts a line. It was verified by putting one of the four guards back and watching it fail, naming both sides of the rename.

The nine catalogues are re-extracted and recompiled. Every file here was renamed by the conversion, so every source reference in them pointed at a path that no longer exists. Nothing translated is lost: the thirty-seven ids retired are strings no longer in the source. Twelve strings arrive that predate this branch, being in main's source but never in main's catalogue.

## Keeping it that way

A `ui-type-check` job runs `tsc --noEmit` beside `ui-lint`, as a required check rather than an advisory one, and `ui-lint` gained the catalogue check described above. `awx/ui/CONTRIBUTING.md` gains a section on how props are typed, what to reach for in place of `any` and where the API types come from, in place of the PropTypes section describing a library the UI no longer carries.

The test files come under eslint with the same rules as the source, less the handful that describe an application rather than a test of one: an assertion names the string the user sees, a test file has no default export to be its subject, and a stub component reads whatever the real one is given. That turned up nineteen unused imports, a component stub calling a hook from a function named `default`, a catalogue imported by a file name nothing else uses, and an entry imported as `./index.js` from a directory with no such file.

## Verified locally

```
npm --prefix awx/ui run type-check    # clean
npm --prefix awx/ui run lint          # clean
npm --prefix awx/ui run prettier-check
npm --prefix awx/ui run test          # 551 files, 2991 tests
npm --prefix awx/ui run build
```

Both of the tests that used to fail about one run in five are fixed rather than tolerated. The settings redirect is a `<Navigate>`, applied in an effect, and the test read the location as soon as the loading state cleared. The node modal was not a test problem at all: choosing a node type set ten formik fields one after another, and the list for the new type mounts between them, so a resource selected in that window was undone by the reset still arriving. A person cannot click inside those microseconds, which is why only the suite saw it. The fields are set together now.


blaipr:
Work here is finished, and the branch is rebased onto main now that #916 has landed.

Verified locally, since CI does not run on pull requests from forks: lint, prettier, the type check, the message catalogue check, both halves of the test suite (551 files, 2991 tests) and a production build all pass.

Ready to merge.

cigamit (approved):

PR #920, opened 2026-09-12
Remove openshift/python-daemon/pexpect packages

cigamit:
* Openshift package is not imported directly, so drop it and just bring in the kubernetes package, which we do need.
* Ansible-Runner finally dropped python-daemon yesterday, which removes it and lockfile
* We also don't use pexpect, so remove our pin.  Its still brought in by Ansible-Runner though.


TheWitness (commented):
Nothing to see here Jimmy.

cigamit:
Heh, since I merged a PR that was based off this one, those changes got pulled in there.  All good.

PR #926, opened 2026-09-12
refactor: DER-encode the receptor hostname without asn1

blaipr:
Sits on top of #925, and through it #924, #923, #919 and #920. Fourth of the series from the "same feature, smaller footprint" list on the [roadmap thread](https://forum.ascender-automation.org/d/15-to-do-list-of-major-things-on-the-roadmap).

## Problem

`asn1` is a dependency for four lines in `generate_receptor_tls`: encoding the instance hostname as a UTF8String so it can be the value of the receptor OID in the certificate's SANs. Nothing else in the tree writes ASN.1.

## Change

A `der_utf8_string` helper in the same module. The encoding is a tag byte, a length and the payload, with the long form handled even though a hostname never reaches 128 bytes.

Worth correcting something I wrote on the roadmap thread: I said `cryptography` already does this in a few lines. It does not. It takes the `OtherName` value as encoded bytes and exposes no general encoder, so the helper is written here rather than borrowed from it.

## Checks

Before removing the package, the helper was compared against it byte for byte on eight inputs: the empty string, one character, a hostname, 127 bytes and 128 bytes either side of the short form boundary, 300 bytes, 65535 bytes, and a multibyte string where the length must count bytes rather than characters. They agree on all eight.

`awx/main/tests/unit/api/test_instance_install_bundle.py` keeps six of those as a parametrised test plus the multibyte case, asserting literal expected bytes so the test does not depend on the package being removed.

`ruff check` and `ruff format --check` clean. `requirements.txt` pruned by provenance rather than recompiled, since the updater pins python3.14 and this machine has 3.12; `asn1` had no dependencies of its own.

## Not tested here

Generating a real install bundle and having receptor accept the certificate. The encoder is verified against what the package produced, so the bytes in the SAN are unchanged, but an end to end bundle is worth generating before release.


blaipr:
Duplicate of #921, which came first and does it better: `cryptography.hazmat.asn1.encode_der` has existed since 47.0, so the hand-rolled encoder here is unnecessary. The PR body's claim that cryptography exposes no encoder was simply wrong, and #921 also lands a test at the same path. Closing in favour of it.

PR #929, opened 2026-09-12
build: install UI dependencies with --ignore-scripts

blaipr:
Sits on top of #928 and the rest of that stack. From the supply chain section of the [roadmap thread](https://forum.ascender-automation.org/d/15-to-do-list-of-major-things-on-the-roadmap).

## Problem

`npm ci` runs each package's install scripts by default, which is how the Axios class of attack executes: a compromised release runs code during install, before anything is built, linted or reviewed. Three places install the UI dependencies and none of them opts out.

## Change

`--ignore-scripts` on all three: the Makefile target that `make ui-release` and therefore the image build reach, the CI job, and `awx/ui/Dockerfile`.

Nothing in this tree needs those scripts. The lockfile marks two packages with `hasInstallScript`:

- `fsevents`, optional and macOS only, so it is never installed on the build hosts.
- `unrs-resolver`, whose native binding is also published as a platform `optionalDependency`, so npm installs the right one without running the script. Its postinstall is the fallback for when that does not happen.

## Checks

Run locally on node 24, which is one of the versions `package.json` allows:

- `npm ci --ignore-scripts` completes and reports 0 vulnerabilities.
- `node_modules/@unrs/resolver-binding-linux-x64-gnu` and `-musl` are present afterwards, so the resolver has its binding.
- `npx eslint src/screens/Login` exits clean, which is the thing that would break first if import resolution were broken.
- `npx vite build` produces the expected tree, `index.html` plus `static/{css,js,media}`, 11 MB.

## Worth knowing

This does not protect against a malicious package that runs at import time rather than at install time, and it does not replace pinning or auditing. It closes the window where install alone is enough.


blaipr:
Thanks, went through all three.

**Applied: the Makefile one.** It was right and it mattered. `--ignore-scripts` covered only the `awx/ui/node_modules` rule, while `ui-lint`, `ui-type-check`, `ui-test`, `ui-test-screens` and `ui-test-general` each ran their own bare `npm --prefix awx/ui install`. Four of those five are CI matrix entries, so an install script could still run on every pull request and the protection was worth less than it looked.

Rather than repeating the flag five times, those targets now depend on `awx/ui/node_modules`, which is the rule that already does the protected install. That is what `ui-devel` and `ui-devel-test` already did, so the file is consistent as well as safe, and there is one place left to change if the install ever needs to differ.

Checked with `make -Bn` on each of the five: one `ci --ignore-scripts` and no bare install, where before each had a bare install and no flag.

**Not applied: the two pnpm ones**, on `.github/workflows/ci.yml` and `awx/ui/Dockerfile`. Two reasons.

Both lines they point at already read `npm ci --ignore-scripts`. That is what this pull request changed them to, so there is no missing flag there.

And this repository does not use pnpm. There is no `pnpm-lock.yaml` and no reference to pnpm anywhere in the tree; the UI is pinned by `awx/ui/package-lock.json`. Switching package manager would be a different change with its own lockfile, its own CI setup and its own review, and doing it inside a supply-chain hardening pull request would make both harder to reason about. Happy to look at it separately if there is an appetite for it.

cigamit (approved):

PR #932, opened 2026-09-12
refactor: remove the Insights credential kind

blaipr:
Sits on top of #931. Second of three, from the "inherited Red Hat surface" section of the [roadmap thread](https://forum.ascender-automation.org/d/15-to-do-list-of-major-things-on-the-roadmap).

## Problem

`insights` is one of the `KIND_CHOICES` on `CredentialType`, and nothing can have it:

- there is no managed Insights credential type, so none ships;
- a custom credential type is restricted to `cloud` or `net` by `serializers.py`, so none can be created;
- the only consumer was the Insights project type, removed in #931.

## Change

The choice, the two inventory source validation messages that listed `insights` among the disallowed kinds, and the entry in the UI's `unacceptableCredentialTypes`, which named a credential namespace that cannot exist.

## No migration, on purpose

Django does not enforce choices at the database level, so a row carrying `kind='insights'` in a database migrated from AWX keeps loading exactly as before. It was already unusable, since the project type that consumed it is gone. Deleting such rows is a decision for whoever runs that install, and a data migration that removed credentials would be a poor default.

## Checks

`ruff check` clean on the three Python files, `npx eslint` clean on `Credential.tsx`. The credential type test parametrises over kinds a custom type may not use, so `insights` leaves that list with the choice itself.

## Left alone deliberately

`awx/ui/src/screens/Credential/shared/data.credentialTypes.json` still contains an Insights row. It is a captured API response used as test data rather than a description of what exists, and editing it would churn tests unrelated to this change.


blaipr:
Amended: this was missing its migration.

Removing `insights` from `KIND_CHOICES` without one leaves Django's autodetector seeing the model and the migrations disagree. `makemigrations --check` fails, `makemigrations` offers a spurious `0215` on every run, and the next real migration anybody generates silently carries this change inside it. My original message argued no migration was needed because PostgreSQL does not enforce choices, which is true about the data and beside the point about the state.

`0215_alter_credentialtype_kind` is in the commit now. It emits no meaningful schema change, and a row carrying `kind='insights'` in a database migrated from AWX still loads exactly as it did.

Verified on a fresh database: migrate applies cleanly through `0215`, and `makemigrations --check main` then reports no changes. The rest of the stack is rebased on it.

cigamit (approved):

PR #936, opened 2026-09-12
test: fail the build on a new accessibility violation

blaipr:
Sits on top of #935 and the rest of that stack. From the "architecture and code shape" section of the [roadmap thread](https://forum.ascender-automation.org/d/15-to-do-list-of-major-things-on-the-roadmap).

## Change

`@axe-core/playwright` scans the rendered page against WCAG 2 A and AA on three screens: the login page a signed out visitor sees, and the jobs and templates lists that most of the application is built from. The spec lives with the other end-to-end tests and `npm test` already runs everything in `tests/`, so the CI workflow needs no change.

It compares what each screen violates against a recorded baseline and fails only on something **new**. A zero violation gate on an application that does not pass WCAG today would fail on arrival and be switched off within a week; this one fails on regressions, which is the thing worth catching.

**No baseline file ships, because there is nothing to record.** All three screens are clean.

## The one thing the scan found, fixed rather than baselined

`awx/ui/index.html` had no `<title>`. `Login.tsx` sets one, but only after its branding request resolves, so until then the document has no title at all: axe reports `document-title`, a screen reader announces nothing useful, and the browser tab shows the URL. The template now ships a static `<title>Ascender</title>`, which the custom branded title still overrides at runtime.

## Checks

Run against a development stack brought up from this branch, `make docker-compose` with the UI built and the web process restarted so it serves it:

- before the fix: login failed with `document-title`, jobs and templates passed;
- after the fix and a UI rebuild: **3 passed**, and `curl` confirms the served `index.html` carries the title.

## Worth knowing

Three screens is a start, not coverage. The forms are where accessibility problems usually live, and they need a signed in session and seeded data, which the existing `seed.js` already provides. Adding them is the obvious follow up once this is green in CI.


blaipr:
Finished this one off, and in doing so found that it did not work.

**It had no baseline.** The gate reads `accessibility-baseline.json` and treats anything not in it as new. The file was never written and never committed, so `baseline()` returned `{}`, every existing violation counted as a regression, and all three tests failed on arrival. It is recorded and checked in now.

**And the wait was wrong, which hid what the gate was for.** `networkidle` fires while React still has an empty root, so axe was scanning 21 elements and reporting nothing: `jobs` looked clean when it has three violations. Waiting for the page shell and letting it settle gives the real answer. `networkidle` is gone rather than kept alongside, because on the two forms that poll it never fires at all and the test times out at 90s.

**Then the follow up this pull request asked for.** Three list screens is not coverage, and the forms are where accessibility problems live, so seven of them are scanned now: project, inventory, credential, organization, user, team and job template.

That found something immediately:

| | |
| - | - |
| every list and form | `button-name`, `link-name` |
| the two lists, additionally | `color-contrast` |
| the job template form, alone | **`label`** |

`label` is a control with nothing naming it, on the biggest form in the application. Three list screens would never have seen it.

Checked by taking `label` back out of the baseline, which fails the gate with exactly what it should say:

```
Error: new accessibility violations on template-add: label
```

The e2e suite is **21 passed**, where it was 14 with 3 of them failing.

cigamit (approved):

PR #937, opened 2026-09-12
ci: fail on a type regression in the packages that are clean

blaipr:
Sits on top of #936 and the rest of that stack. From the "architecture and code shape" section of the [roadmap thread](https://forum.ascender-automation.org/d/15-to-do-list-of-major-things-on-the-roadmap), which asks to promote `ty` from advisory to blocking one package at a time.

## Problem

`ty` runs over the whole tree and reports without failing, which is the right setting while there is a backlog: **2,896 diagnostics** today. The cost is that a package reporting nothing now can start reporting tomorrow and nothing says so.

## Change

`typecheck-strict`: the same pinned checker, failing, over the twelve packages that already report nothing, plus a blocking `api-type-check-strict` matrix entry. The advisory whole-tree run stays exactly as it is, so the backlog is still visible and still not a merge blocker.

A package moves into the strict list the moment it comes clean. That is the "one package at a time" the roadmap asks for, made mechanical.

## Where the backlog actually is

Measured in the devel image, so whoever works it next knows where to start:

| area | diagnostics |
| --- | --- |
| `awx/main/tests` | 1,107 |
| `awx/main/models` | 603 |
| `awx/api/views` | 144 |
| `awx/api/serializers` | 112 |
| `awx/sso/tests` | 92 |
| `awx/main/credential_plugins` | 89 |
| `awx/main/tasks` | 79 |

## Checks

Run in a devel image built from this branch: `tox -e typecheck-strict` reports **All checks passed**, and the advisory `typecheck` still exits 1 with the 2,896 diagnostics, unchanged.


blaipr:
Rebased onto current `main` and this is mergeable again.

cigamit (approved):

PR #940, opened 2026-09-12
refactor: rename TOWER_URL_BASE to ASCENDER_URL_BASE

blaipr:
Sits on top of #939 and the rest of that stack. From the "inherited Red Hat surface" section of the [roadmap thread](https://forum.ascender-automation.org/d/15-to-do-list-of-major-things-on-the-roadmap): rename the `TOWER_*` settings, `TOWER_URL_BASE` most of all, keeping aliases for a release.

## Change

`TOWER_URL_BASE` becomes `ASCENDER_URL_BASE`: the registered setting, the 56 references across the API, the analytics collectors, the notification templates, the log formatter, the UI settings screen and the docs.

Conf migration `0011` carries the stored value across, using the same `rename_setting` helper as the three earlier renames in this repo. A deployment that sets the old name in `/etc/tower/conf.d` keeps working for a release through a fallback in `production.py`; the installers set it through the API, which the migration covers.

Two more in the same sweep:

- `TOWER_SECRET_KEY`, the environment variable `regenerate_secret_key` reads, becomes `ASCENDER_SECRET_KEY`, with the old name still honoured.
- The Ascender credential type now injects `ASCENDER_HOST`, `ASCENDER_USERNAME`, `ASCENDER_PASSWORD`, `ASCENDER_VERIFY_SSL` and `ASCENDER_OAUTH_TOKEN` beside the `TOWER_` and `CONTROLLER_` names it already sets, so ascender-kit and the collection have something to move to rather than a cliff.

## Left alone deliberately

`LOG_AGGREGATOR_TOWER_UUID` and the `tower_uuid` field in the log payload. That field is a wire format: [ascender-ledger](https://github.com/ctrliq/ascender-ledger) and any other aggregator parse it, so renaming it belongs with the logging work, not here.

## Checks

Against a running development deployment built from this branch, not in theory:

1. seeded `TOWER_URL_BASE = https://old-deployment.example.com` in the database;
2. `awx-manage migrate conf` → OK;
3. the value is now under `ASCENDER_URL_BASE`, and the `TOWER_URL_BASE` row is gone;
4. `GET /api/v2/settings/system/` serves `ASCENDER_URL_BASE` with that value and no longer offers the old name.

Test suites: 1,368 unit and functional tests including `test_settings.py`, the secret key regeneration commands and the credential plugins; the UI Setting screens, 332 tests. The generated API types were regenerated rather than hand edited, and regenerating twice is byte identical, so #934's drift check stays green.

## Follow-ups in other repositories

`ascender-install`, `ascender-k8s-install` and `ascender-pro-install` each set `TOWER_URL_BASE` through the API, and `ascender-collection` names it in a test config. They keep working until the migration runs, and then need the new key; I will open those separately.


blaipr:
This branch had a real failure in it, found while chasing the functional suite to zero.

`test_inventory_update_injected_content[ascender]` failed because **this pull request caused it**: the ascender credential injector gained `ASCENDER_HOST`, `ASCENDER_USERNAME`, `ASCENDER_PASSWORD`, `ASCENDER_VERIFY_SSL` and `ASCENDER_OAUTH_TOKEN` alongside the `TOWER_` and `CONTROLLER_` ones, and the reference environment the test compares against still listed ten variables where the injector now produces fifteen.

```
AssertionError: assert {'TOWER_HOST'...} == {'TOWER_HOST'...}
  Right contains 5 more items:
  {'ASCENDER_HOST': 'https://foo.invalid', 'ASCENDER_OAUTH_TOKEN': '', ...}
```

`awx/main/tests/data/inventory/plugins/ascender/env.json` now records all fifteen, in the shape the test's own generator writes: indent 4, keys sorted.

## Why this was not visible before

The functional suite reports 96 failures on a reused test database, and this was one of them, indistinguishable from the noise. `pytest.ini` passes `--reuse-db`, so the database survives between runs; the coverage target passes `--create-db` and builds it fresh, which is why CI is green and only a local `make test` sees them.

Rebuilding the database takes the suite from 96 failures to one, and that one was this.

## Where it stands now, on this branch

| | |
| - | - |
| `awx/main/tests/functional` | **2,338 passed, 5 skipped, 0 failed** |
| `awx/main/tests/unit`, `awx/conf`, `awx/sso/tests` | **1,691 passed, 1 skipped** |

I also spent a while on a fix for the other 95 that I then threw away, which is worth recording so nobody repeats it: I diagnosed them as leftover hourly event partitions holding foreign keys that make `TRUNCATE` fail, and wrote a conftest fixture to drop them. Checking whether it actually dropped anything said no: `main_jobevent` is not partitioned in the test database at all, because `--nomigrations` builds the schema from the models. The partitions are in the development database, which is where I had looked. The 95 are a test database whose schema predates something, and `--create-db` is what resolves them.

cigamit (approved):

PR #942, opened 2026-09-12
fix: give the statement timeout a source that is not uwsgi

blaipr:
Sits on top of #941 and the rest of that stack. Step one of [Path to ASGI](https://forum.ascender-automation.org/d/61-path-to-asgi), and the one step there worth doing whether or not the rest happens.

## Problem

`awx/settings/statement_timeout.py` imports `uwsgi` and derives PostgreSQL's `statement_timeout` from the `harakiri` value, so a query is cancelled before uwsgi kills the worker. Off uwsgi that import fails and it falls back to `DATABASE_STATEMENT_TIMEOUT`, which is `None` by default, so **no timeout is applied at all**.

That is invisible today, because uwsgi is what serves. It stops being invisible the moment anything else does: daphne or uvicorn would serve HTTP with no statement timeout anywhere, and nothing in the tree would report it.

## Change

A process that announces itself with `AWX_WEB_PROCESS` gets `DATABASE_STATEMENT_TIMEOUT`, or `110000` ms when that is unset, which is exactly what the uwsgi path works out to today: harakiri of 115 seconds less the five second margin. The web supervisor programs, both `uwsgi` and `daphne`, set the variable.

Everything else is untouched: task workers, management commands and migrations carry no marker, so they still get no timeout, because a long query there is the job rather than a symptom. And a uwsgi deployment behaves exactly as before, since harakiri is read first and still wins.

## Checks

- `awx/main/tests/unit/settings`: 35 tests, including three new ones for the web marker, the setting taking precedence over the default, and a task process being left alone.
- Demonstrated in the devel image: with the marker, `-c statement_timeout=110000`; without it, no options are added at all.

---

# Second commit: let the ASGI application serve HTTP, not just websockets

Collapsing onto one ASGI server needs that server to be able to answer the API. It cannot today:

```
$ curl -u admin:password http://127.0.0.1:8051/api/v2/ping/
<h1>500 Internal Server Error</h1>
ValueError: No application configured for scope type 'http'
```

## Three things were in the way

**Channels 4 no longer fills in an http application.** Channels 2 did when the protocol mapping left it out. `ProtocolTypeRouter` now raises instead, so the ASGI application serves websockets and nothing else. The mapping names one.

**Seven call sites read `request.environ`.** That attribute is on `WSGIRequest` and not on `ASGIRequest`, so every request died with `AttributeError: 'ASGIRequest' object has no attribute 'environ'`. Under WSGI, `self.META = environ` is the same dict object, and under ASGI Django builds `META` from the scope with `REMOTE_ADDR`, `REMOTE_HOST` and the headers these sites want. They read `META`, which is identical under WSGI and correct under ASGI.

**`URLModificationMiddleware` registered two settings in `__init__`.** Django builds the middleware chain once per handler, and there is more than one handler the moment both applications exist, so the second raised `ImproperlyConfigured: Setting "NAMED_URL_FORMATS" is already registered.` It now registers once per process.

## The ASGI application now answers the API

53 endpoints fetched through both servers and compared:

| | |
| - | - |
| byte-identical status and body | **51** |
| different | **2** |

Both differences are noise rather than behaviour. `/api/v2/bulk/` is the browsable page, which differs from **itself** on 14 lines between two calls to the same server: CSRF token, request id, timings. `/api/v2/metrics/` returns the same four `awx_status_total` values in a different line order.

## What this deliberately does not do, and why

It does not switch anything over, because the measurement says which server the collapse would have to use. Through nginx, 60 requests at each concurrency:

**`/api/v2/ping/`**

- 1 at a time: uwsgi 78ms, daphne 85ms, uvicorn 84ms.
- 5 concurrent: uwsgi **41.1 req/s**, daphne **7.5**, uvicorn **19.3**.
- 20 concurrent: uwsgi **45.2 req/s**, daphne **6.9**, uvicorn **34.2**.

**`/api/v2/job_templates/`**

- 1 at a time: uwsgi 497ms, daphne 509ms, uvicorn 481ms.
- 5 concurrent: uwsgi **4.6 req/s**, daphne **1.5**, uvicorn **3.3**.
- 20 concurrent: uwsgi **4.6 req/s**, daphne **1.4**, uvicorn **4.9**.

One request at a time the three are indistinguishable. Under load daphne is three to six times worse, because it is one process and one event loop running sync views behind the GIL, against five uwsgi worker processes. **daphne is not a replacement for uwsgi**, and a collapse that assumed it was would have cost most of the throughput.

uvicorn with five workers is a replacement, and on the heavier endpoint at 20 concurrent it is slightly ahead. That is the shape the next step should take, and it is a dependency change rather than a one-line one, so it is not in here.

## Checks

- Three tests, **two of which fail on the commit before this one**: the mapping answers `http`, the named url middleware works on an `ASGIRequest`, and no module under `awx/` reads `request.environ`. The third catches the reintroduction of any of the seven.
- Live on the development stack after the change: the API through nginx and uwsgi still 200, named url rewriting still resolves `/api/v2/organizations/Default/` and still 404s an unknown one, and daphne answers `/api/v2/ping/`.
- `awx/main/tests/unit`, `functional/api`, `test_routing` and `test_named_url`: the same **8** failures before and after.

---

---

# The uvicorn swap has moved to #975

This pull request and that one were the same change, and only one of them is a judgement call.

**What stays here** is the part with no trade in it: the statement timeout off uwsgi, the ASGI application able to answer HTTP, and a test holding the logs to stdout. Nothing about how a deployment serves traffic changes.

**What moved to [#975](https://github.com/ctrliq/ascender/pull/975)** is replacing uwsgi and daphne with one uvicorn, which is at parity on the heavier endpoint under load and about half on the trivial one, with no later payback now that async views have been measured. That wants deciding rather than reviewing, and it was holding up three commits that do not.

#975 sits on this one, so the order is: this, then that if it is wanted.


blaipr:
Looked at the rest of this bullet, "one process per container: logs on stdout, no supervisor tree", and the two halves are in very different places.

**Logs on stdout is already done.** `AWX_LOGGING_MODE` decides it: `file` writes the nine per process logs to `/var/log/tower`, which is what a VM install wants, and `stdout` nulls those handlers and opens the console so the runtime collects them. Both the Dockerfile and the compose file set `stdout`, and the running container confirms it. Nothing to change.

What was not guarded is that it stays true. Every handler in `handler_config` goes through one loop that nulls the file class under stdout, but a handler added straight to `LOGGING['handlers']` with a `filename` would skip that loop and write a file inside a container that nobody reads. Four tests now load the settings under each mode and hold it: nothing writes a file under stdout, the nine still do under file, the console is gated behind debug on a VM and open in a container, and a mode that is neither is rejected rather than quietly treated as one of them.

**One process per container is a deployment change, and I have stopped short of it deliberately.** The image already splits into three: web runs nginx, uvicorn, ws-heartbeat and the cache clear loop; task runs the dispatcher, wsrelay and the callback receiver; rsyslog runs two. Going further means moving nginx and the two small loops out into their own containers in the operator's templates, and this repository cannot show that working. It wants doing where it can be deployed and watched, not asserted here.

So what this pull request now carries for the item is: the ASGI server serving both protocols, uwsgi and daphne both gone, and the logging half pinned. The container split is the piece left, and it belongs in `ascender-operator`.

blaipr:
I have been saying, here and in the commit message, that this pull request's throughput cost is "exactly what the async views item removes". I went to measure that before recommending the two be decided together, and it is **wrong**.

## What I measured

Two views doing identical work, a twenty row ORM read, one `def` and one `async def`, plus a third using Django's own async ORM API. Served by uvicorn, through nginx, `ATOMIC_REQUESTS` off so the async ones can run at all:

| concurrency | sync | async | async ORM |
| - | - | - | - |
| 1 | **9.3 req/s**, 98ms | 7.7 req/s, 120ms | 8.0 req/s, 122ms |
| 5 | 11.8 req/s | **12.5 req/s** | 11.9 req/s |
| 20 | **19.2 req/s** | 18.1 req/s | 17.1 req/s |

Async is not faster. It is about 20% slower at one at a time and level within noise under load.

## Why, and why it will stay that way

The ORM is synchronous. An `async def` view that reads the database still crosses into a thread to do it, so the hop does not go away, it moves from the view boundary to each ORM call. `async for obj in Model.objects.all()` is the same threadpool underneath, which is why the third column is no better than the second.

Async views pay off for a view that waits on something that is already async, and AWX views wait on the database.

## And the combination is refused outright

Not a preference, a `RuntimeError`:

```
RuntimeError: You cannot use ATOMIC_REQUESTS with async views.
```

So the two are not a package. Turning `ATOMIC_REQUESTS` off buys nothing here while the ORM is sync, and it gives up a transaction per request to buy it.

## What that means for this pull request

It has to stand on its own, which changes the question. It is not a down payment on a throughput win that arrives later; it is one server instead of two, at parity on `/api/v2/job_templates/` under load and about half on `/api/v2/ping/`. Worth having for the operational simplicity, or not worth the trivial-endpoint cost. That is a judgement rather than a measurement, and it is not mine to make.

What I would not do is merge it *expecting* async views to pay it back.

cigamit (approved):

PR #943, opened 2026-09-12
refactor: vendor the ascii plot instead of installing it

blaipr:
Sits on top of the stack ending at #942. From the roadmap list in [To Do](https://forum.ascender-automation.org/d/15-to-do-list-of-major-things-on-the-roadmap): "`asciichartpy`: keep `graph_jobs`, vendor the plot function or keep the package, the weakest item." This is the vendor.

## What it does

`asciichartpy` ships on every production image for one debugging management command, `graph_jobs`, which plots pending, waiting and running jobs on a terminal. The package is a single MIT licensed file with no dependencies of its own, and its `plot` function has not changed in years.

The function moves in tree as `awx/main/utils/ascii_chart.py`, with the doctests and the `from __future__ import division` dropped and nothing else touched. `licenses/asciichartpy.txt` stays where it is, and the module header points at it and at [kroitor/asciichart](https://github.com/kroitor/asciichart).

## Same output, checked rather than asserted

The vendored copy and the installed package were run against 306 series: the six documented examples, and 300 random ones with NaN gaps, one to three series, random heights and colours. They agreed **character for character on every one**, and all nineteen exported names are identical.

## Checks

- `awx/main/tests/unit/utils/test_ascii_chart.py`: nine new tests pinning the documented cases to the strings the package itself produced, so a change to the vendored copy that alters a single character fails here.
- `awx/main/tests/unit/utils`: 210 passed.
- `graph_jobs` imports and plots against the vendored module, verified in the devel image.


blaipr:
Closing this one: on reflection the trade is not worth it.

What it bought was one fewer PyPI package on every production image. What it cost was 120 lines of someone else's MIT code that we then own, review and keep, for a debugging management command. That is the wrong direction for a tree that is trying to get smaller, and `asciichartpy` is a single file with no dependencies of its own, so it was never the part of the footprint worth attacking.

PR #944, opened 2026-09-12
refactor: replace ace with CodeMirror 6 in the code editor

blaipr:
Sits on top of the stack ending at #942. From the roadmap list in [To Do](https://forum.ascender-automation.org/d/15-to-do-list-of-major-things-on-the-roadmap): "Swap `ace-builds` for CodeMirror 6: one non-test file imports it, and CodeMirror ships only what is used."

## Why

`CodeEditor.tsx` is that one file, and ace is the largest thing the UI ships. `ace-builds` arrives whole, and its modes and its theme are separate files imported for their side effects, so nothing can be dropped.

## What it costs the bundle

| | before | after |
| - | - | - |
| the CodeEditor chunk | 699.78 kB (193.77 kB gzip) | **433.07 kB (141.01 kB gzip)** |
| all of `static/js` | 4,735,923 bytes | **4,468,204 bytes** |

It was the biggest chunk in the application and it is now smaller than `Config`. Package count goes the other way, two packages out and nine CodeMirror and lezer ones in, which is the shape of that library: the win is the bytes that reach a browser, not the row count in `package.json`.

## What does not change

The component's props are the same, so no caller changes: the same four modes, the same auto height between `minRows` and `maxRows`, the same Enter to edit and Escape to leave, the same 250 ms debounce, the same Ctrl-F search box.

The colours are ace's twilight palette **to the hex**, but they are now CSS variables with those values as fallbacks. The four themes set the variables on `.cm-editor` rather than overriding rules with `!important` to reach ace's classes, which is what `dark.css` and `default.css` were doing.

## What changes for the better

- The editor renders only the lines in view, where ace built a line of DOM per line of content. That is the cost `MAX_ROWS` exists to cap, and a host's facts run to thousands of lines.
- A read-only editor is now a tab stop. Its scroll area was otherwise unreachable from a keyboard, which axe reports as `scrollable-region-focusable`: found while checking this change in a browser, and fixed here.
- The editable content carries `aria-labelledby` pointing at the PatternFly label that names the field, so the editor announces itself the way the old hidden textarea did.

## Checks

**Unit**: 551 files, 2,998 tests, all passing. `tsc --noEmit` and `eslint src` clean.

**The tests got better**, which is worth its own line. The editor keeps its document in the DOM, one element per line, so what it shows and what is typed into it are both observable. Three test files stop mocking the editor away, and the assertions that were dropped when the suite moved to RTL come back: formatted JSON, yaml to json and back, the `---` and `{}` empty-value defaults, and the type-then-submit path through formik.

**In a real browser**, against the development stack:

- typing into an inventory's Variables field: `---` then `---\nfoo: bar`, toggled to `{"foo": "bar"}` and back, which only works if `onChange` reached formik
- Ctrl-F opens the search panel, Escape closes it and leaves focus in the editor
- a read-only editor on Settings: `contenteditable=false`, `tabindex=0`, no active-line highlight, named by its label
- background `#13161b` and gutter `#1a1e25`, the values the dark theme sets, so the theme variables land
- no console errors on any of it

**End to end**: 12 of 14 pass. The two that fail are `accessibility.spec.js` on `/jobs` and `/templates`, and they are not this change: neither screen mounts an editor. Both report `button-name` and `link-name` on `#toolbar-workflow-approval-badge`, the pending-approvals badge in the page header, which renders only when approvals are pending, and the e2e seed creates some. A `color-contrast` on the Finish Time column header comes with it. Real bugs, pre-existing, and a follow-up of their own.


blaipr:
Amended: the licenses were missing.

`awx/main/tests/functional/test_licenses.py` requires a file under `licenses/ui/` for every production dependency in `package.json`, and fails when one is absent or when a file has no dependency behind it. This swap did both: nine new packages with no license file, and two files left behind for packages it removed.

Added `codemirror-{state,view,commands,search,language,lang-json,lang-yaml,legacy-modes}.txt` and `lezer-highlight.txt`, all MIT, taken from the packages themselves. Removed `ace-builds.txt` and `react-ace.txt`.

`test_python_and_js_licenses` passes. Found while running the backend suite for a later change in the stack, which is rebased on this.

cigamit:
Nice, seems like we are coming full circle after this
https://github.com/ansible/awx/pull/9281

cigamit (approved):

blaipr:
Full circle indeed, and worth taking seriously: that PR closed three issues, so this one owes an answer to each rather than rediscovering them.

**ansible/awx#8905**, "explore replacing react-codemirror2", was the actual driver: the *wrapper* was heading for abandonment, not CodeMirror. There is no wrapper here. CodeMirror 6 ships as `@codemirror/state`, `@codemirror/view` and the language packages, all first party, and the React integration is about forty lines in our own component.

**ansible/awx#8176**, tab being swallowed instead of moving to the next field, was already answered: the keymap binds no tab, and a focus wrapper holds the tab stop, so tab moves on, Enter enters the editor and Escape or Shift-Tab leaves it. That is the same pattern keithjgrant described in #9281, and the helper text under the editor says so.

**ansible/awx#8827**, tabbing into YAML landing on the `---` instead of the first content line, was still open here. A fresh CodeMirror state starts its selection at the top of the document, so the first keystroke went into the syntax. Just pushed: the opening cursor sits below the marker, at the end of it where the marker is the whole document, and at the top where there is no marker to skip.

Three tests cover it, and they fail against the editor as it was a commit ago: typing `x` gave `x---`, not `---x`.

PR #946, opened 2026-09-12
perf: let a web process reuse its database connection

blaipr:
Sits on top of the stack ending at #945. From the roadmap list in [To Do](https://forum.ascender-automation.org/d/15-to-do-list-of-major-things-on-the-roadmap): "No pooling on the database: `ATOMIC_REQUESTS` is on, `CONN_MAX_AGE` is unset and psycopg's pool is off."

## The measurement first

`CONN_MAX_AGE` is unset, which is Django's default of `0`: **every request opens a PostgreSQL connection and closes it** when the response is sent. Nothing pools, so that is the cost of every API call before it has run a single query.

| | |
| - | - |
| 50 requests to `/api/v2/ping/`, today | **51** database sessions |
| establishing one connection | **5.20 ms** |
| the same with the `application_name` and `statement_timeout` options in it | **5.68 ms** |
| a query on a connection that already exists | **0.29 ms** |

Both numbers came off the development stack and its PostgreSQL, the session count from `pg_stat_database.sessions` either side of the fifty requests.

## The change

A process that serves HTTP keeps its connection for sixty seconds. The same fifty requests then open **5** sessions, one per uwsgi worker: a 90% cut in connection churn, for no change anyone can see.

Sixty seconds is deliberately short. An idle worker holds its connection until its next request notices the age, so what is held at rest is one per web worker, and a short life keeps a connection from outliving a network path that has quietly gone away.

Task workers, management commands and migrations are left alone: they hold one connection for the life of a long process, or want it closed the moment the command ends, and an age limit helps neither. `DATABASE_CONN_MAX_AGE` overrides all of it, including for a task process, so a deployment can turn reuse off with `0` or hold connections for longer.

Web is detected the way #942 detects it: `uwsgi` answers for every deployment today, and `AWX_WEB_PROCESS` answers for anything that is not uwsgi, which is what will matter if the server ever becomes daphne.

## Why not psycopg's pool

A pool helps a process handling several requests at once. Under uwsgi's prefork model each worker handles one at a time, so a pool per worker would hold the same single connection with an extra dependency, `psycopg_pool`, to do it. The pool becomes the right answer when the server is async, which is the [Path to ASGI](https://forum.ascender-automation.org/d/61-path-to-asgi) thread, not this.

## Checks

- `awx/main/tests/unit/settings`: 49 passing, 11 of them new, covering the marker, uwsgi, the setting winning over the default, `0` turning reuse off, a task process being left alone, and the empty `DATABASES` cases.
- Settings imported as each kind of process in the devel image: task gets `CONN_MAX_AGE=0`, web gets `60` alongside the statement timeout.


cigamit (dismissed):

blaipr:
Conflict resolved, this is mergeable again.

cigamit (approved):

PR #947, opened 2026-09-12
perf: read one key to decide the settings cache is stale

blaipr:
Sits on top of the stack ending at #946. From the roadmap list in [To Do](https://forum.ascender-automation.org/d/15-to-do-list-of-major-things-on-the-roadmap), the third clause of "Type the configuration: 1,302 lines of defaults, 78 mutable settings, **a cache per request**".

## What it does today

`SettingsCacheMiddleware` clears this process's in-memory settings cache at the start of **every** request. The cache therefore never survives one, so every setting read during the request goes to the shared cache instead.

It is there for a reason, and the docstring says so: a script can POST to `/api/v2/settings/all/` and immediately GET it back, and expect the new value even when the two requests land on different workers. The price is paid by every request, to cover a setting write that almost never happens.

## The measurement

200 requests each way against the development stack, with background cache traffic measured over the same window and subtracted:

| endpoint | before | after |
| - | - | - |
| `/api/v2/ping/` | 7.11 round trips per request | **2.10** |
| `/api/v2/me/` | 12.45 round trips per request | **5.38** |

## The change

Every setting change bumps one shared counter. The middleware reads that one key and drops the in-memory cache only when it has moved since this process last looked.

`handle_setting_change` already deleted the changed keys from the shared cache and cleared its own in-memory copy; it now also bumps the counter, which is what tells the other workers.

## Read-your-writes is unchanged, and checked

With five uwsgi workers warmed so each held its own in-memory copy:

- the counter key was **absent** before the write
- a `PATCH` to `/api/v2/settings/ui/` moved it to **1**
- **twenty immediate reads**, spread across those five workers, all returned the new value

## Failure behaviour

A cache that cannot be read drops the in-memory copy and logs it. That is what the old code did on every request anyway, so an unreachable cache degrades to today's behaviour rather than to stale settings. A cache that has lost the counter looks like a change to every process, which costs one dropped cache each and is the safe direction.

## Checks

- `awx/main/tests/unit/test_settings_cache_middleware.py`: six new tests covering the first request, an unchanged counter, a moved counter, that exactly one key is read, an unreachable cache, and an empty cache being a value rather than "never looked".
- `awx/main/tests/unit` and `awx/conf/tests`: 1,432 passed, 1 skipped.

---

# Second commit: check that every setting's default is one its own field accepts

Another clause of the same roadmap item, so it goes here rather than in a pull request of its own.

## What was missing

Every registered setting declares a field and carries a default. Nothing ever asked whether the two agree. A default its own field refuses is not a startup failure, it is a 500 the first time somebody reads that setting through the API, a long way from the line that caused it.

## What this adds

`awx-manage check_settings` runs every default through its own field and exits non-zero on a disagreement. The three test targets in the `Makefile` call it beside `check_migrations`, so it runs where that already runs.

```
$ awx-manage check_settings
Every registered setting has a default its field accepts.
```

Today it is a ratchet rather than a fix, which is worth saying plainly:

| | |
| - | - |
| settings registered | **289** |
| carry a default | **238** |
| accepted by their own field | **238** |
| refused | **0** |

## 24 settings are deliberately left out

Their fields declare `depends_on` and judge a value against what sibling settings hold at the time, so their default is only right or wrong alongside the rest of the configuration rather than on its own.

`AUTH_LDAP_GROUP_TYPE_PARAMS` is the example. `LDAPGroupTypeParamsField` checks its keys against the arguments of whatever class `AUTH_LDAP_GROUP_TYPE` names, so judging it alone reports whatever that setting was last set to. I found this by checking them anyway and watching it fail depending on what had run before it. That is pinned by a test rather than left as a comment.

## Checks

- Six tests. Two register a bad default on purpose, an `IntegerField` holding a string and a `StringListField` holding a string, and assert that both the function and the command report it. One asserts the registry is not empty, so the sweep cannot pass by walking nothing.
- `awx/conf`: 119 passed. With `functional/api/test_settings.py`: 158 passed.

---

# Third commit: fail when the code reads a setting nothing defines

A `settings.SOME_NAME` that no settings file defines and no registration carries raises `AttributeError` the first time that line runs. That is the same shape of fault as a view calling a method its model does not have, and nothing checked for it.

Reading the tree for every `settings.NAME` and asking whether it resolves makes a typo or a removed setting a failing test rather than a 500 in front of whoever hits that path first.

**160 distinct settings** are read this way, and all of them resolve today, so this is a ratchet rather than a fix. A deliberately misspelled `JOB_EVENT_RETENTOIN_DAYS` fails it with the file and line that reads it:

```
JOB_EVENT_RETENTOIN_DAYS: main/signals.py:641
```

## What it turned up

Three named url settings are not declared anywhere. They appear when `URLModificationMiddleware` is constructed:

- `NAMED_URL_FORMATS` and `NAMED_URL_GRAPH_NODES` it **registers**, so the registry answers for them.
- `NAMED_URL_MAPPINGS` it only **assigns**, so nothing but a built middleware chain makes `settings.NAMED_URL_MAPPINGS` resolve at all.

That is the opposite of configuration loaded once, and it is now pinned by a test rather than left to be discovered.

## Why this rather than annotating defaults.py

That is the other reading of the roadmap bullet, and I measured it before writing anything: **310 assignments** in `defaults.py`, **0** carrying a type annotation. Adding them buys nothing today.

`settings.X` goes through `LazySettings.__getattr__`, and `ty`, the checker this repository runs and #937 is promoting to blocking, has no Django settings support. So no read site gets checked either way, whether or not the definition is annotated. Typing those 160 reads needs a second type checker alongside `ty`, which is a decision rather than a commit, so I have not made it here.

I also checked whether the definitions disagree with each other, since that would have been a real defect: **11 settings are literal in more than one settings file, and 0 of them change type between files.**

## Checks

- `awx/conf`: **121 passed**.


blaipr:
Conflict resolved, this is mergeable again.

cigamit (approved):

PR #950, opened 2026-09-12
refactor: say read only once instead of on every class that is

blaipr:
Sits on top of the stack ending at #949. A first slice of the roadmap item in [To Do](https://forum.ascender-automation.org/d/15-to-do-list-of-major-things-on-the-roadmap): "Declarative permissions instead of 3,103 lines of per-model access classes, one place to change."

Not the whole item, but complete for the property it names. This takes the safest property in the file, read only, and states it once instead of eight times, on something where a mistake cannot widen anyone's access.

## What it does

Eight access classes are read only: job events, ad hoc command events, project and inventory update events, system job events, job host summaries, the activity stream and workflow approval votes. Each said so by repeating the same three methods.

That is 54 lines of the same thing, and worse, "nobody can write this" is something you learn by reading three method bodies rather than something the class states. `ReadOnlyAccess` says it once, and each class then carries only what actually differs about it: its model, and which of its rows a user may see.

## Two things worth pointing at

**`ReceptorAddressAccess` looks like a ninth and is not.** Its three methods carry `@check_superuser`, so a superuser *can* manage receptor addresses and only everyone else is refused. Folding it in would have taken that away quietly. It is left exactly as it was.

**The mixin goes before `BaseAccess` on purpose.** `access_registry` is built by walking `BaseAccess.__subclasses__()`, which sees direct subclasses only, so a class that stopped being one would silently drop out of the registry and lose its access rules entirely. I hit that on the first attempt and the activity stream tests caught it. `UnpartitionedJobEventAccess` is registered by hand two lines below the walk for exactly the same reason, which suggests it has caught someone before.

## Nothing changes, and it is checked

Every access class was asked `can_add`, `can_change` and `can_delete`, as a plain user and as a superuser, with the module imported before and after:

- **48 classes, 288 answers, zero differences.**
- The access registry maps the same **49 models to the same 49 classes**.

Plus the suites: `awx/main/tests/unit` 1,332 passed and 1 skipped, and every functional test matching rbac, access or activity, **439 passed**.

## Read only is now exhaustive

Worth stating plainly, because "eight of fifty-one classes" reads like a fraction of a sweep and is not one. Read only here means refusing a write to everyone, superusers included. `BaseAccess` answers `can_add`, `can_change` and `can_delete` with `self.user.is_superuser`, so a class that simply omits those methods is superuser-writable rather than read only, and folding it in would take access away rather than restate it.

Asking every class in the registry, as a superuser, which writes it permits:

- **50 access classes, 41 let a superuser write**, so they are not read only and are not candidates.
- **9 carry `ReadOnlyAccess`**, the eight here plus `UnpartitionedJobEventAccess`, which inherits it from `JobEventAccess`.
- **0 refuse a superuser while staying off the mixin.**

So there is no tenth class waiting, and nothing further to do for this property. The other 41 are a different problem, the one the roadmap item is really about: per-model classes expressing rules in code rather than as data. That stays open, and is not made smaller by this.

80 lines out, 32 in.

---

# Second commit: say which object a read follows instead of writing the query

#950 took the safest property, read only, and said it once. This takes the largest one.

## Whether "rules as data" is even the right framing

I went looking before writing anything, because #950 found only nine of fifty classes shared a property and that reads like the idea not fitting. Classifying all 145 write rules by shape:

| | | |
| - | - | - |
| constant, including superuser only | **25** | 55 lines |
| delegation to another access class | **4** | 9 lines |
| a single expression | **35** | 91 lines |
| branching, raising, inspecting the payload | **81** | 721 lines |

So on the write side the answer is no, and not marginally: 82% of the lines are imperative, and of those, 55% branch on the proposed new state rather than on the user and the object, 16% raise a specific message instead of returning a boolean, and 16% consult the licence. A permissions table maps subject, object and action to a verdict. These also take the payload and answer with a reason. **That is a different shape, and the roadmap bullet as written overstates what is available.**

The read side is the opposite. Of 42 `filtered_queryset` definitions the median body is **one line**, and twenty of them are the same sentence: show the rows reachable from an object this user may read. They differ only in which model carries the `read_role` and the ORM path to it.

## What this does

Nineteen of those twenty now name the two things instead of writing the query:

```python
class HostAccess(BaseAccess):
    model = Host
    read_via = (Inventory, 'inventory')
```

`ReceptorAddressAccess` is the twentieth and is deliberately left out, because its rule is not this one. See #973.

## Why it is worth doing, which is not the 38 lines

`ReceptorAddressAccess` is the whole argument. It wrote its filter by hand against `Instance`, which is not a `ResourceMixin` and has no `accessible_pk_qs`, so it raised `AttributeError` for every user `get_queryset()` did not short circuit: everyone who is neither superuser nor system auditor. It has been that way since the 23.8.0 upstream sync in February 2024, and nothing caught it, because there were nineteen other hand-written copies and no shared place to check.

`TestReadVia` asks every declaration whether its model carries the role and whether Django can follow its path. Declaring the broken receptor address rule fails it:

```
AssertionError: ReceptorAddressAccess: Instance has no role accessor
```

## Nothing changes, and it is checked rather than argued

Every access class in the registry was asked for its read queryset as a plain user, an org admin, a system auditor and a superuser, and the SQL compared either side of the change:

- **196 signatures, 192 byte-identical.**
- **4 differ**, all `WorkflowJobTemplateNodeAccess`, which passed `accessible_objects` into an `__in` where the other eleven paths pass `accessible_pk_qs`. The shared path uses the latter, which drops one subquery level through `main_workflowjobtemplate`.

For those four, the row sets were compared directly instead: three users, four templates, several different roles, and then a template deleted with its roles left behind, which is the one case where the extra subquery level could have mattered. **6 comparisons, 6 identical.**

## One thing removed on the way

`InventoryAccess.filtered_queryset` took `allowed` and `ad_hoc` arguments that nothing in the tree passes and the body never read.

## Checks

- `awx/main/tests/unit/test_access.py`: 42 passed.
- `awx/main/tests/unit` and `awx/main/tests/functional` matching rbac, access, activity, inventory, workflow or credential: the same **43** failures before and after.


cigamit (dismissed):

blaipr:

cigamit (approved):

PR #951, opened 2026-09-12
refactor: give every resource its own views and serializers module

blaipr:
Sits on the stack. A roadmap item from [To Do](https://forum.ascender-automation.org/d/15-to-do-list-of-major-things-on-the-roadmap): the API's two biggest modules were one file each, and neither had an order a reader could use.

## What it does

**`awx/api/views/__init__.py`: 4,780 lines and 303 classes, now 793 and 18.** The 285 classes that left are in 19 modules named for the urls file that routes to them, which is the grouping the API already has: workflow, project, job_template, inventory_source, job, credential, host, instance, user, system_job, group, team, ad_hoc_command, schedule, oauth2, notification, role, execution_environment and activity_stream. `inventory.py` and `organization.py` already existed and gained the views that belonged with them.

**`awx/api/serializers.py`: 6,557 lines and 127 classes, now a package of 21 modules.** Same names as the views modules, so the two trees line up. The largest is `base.py` at 1,136 lines, then job at 832, inventory at 660 and workflow at 612.

What stays in `views/__init__.py` is what is not a resource: the shared bases and mixins, the ANSI helpers, the exception handler, and the dashboard and auth views that answer for the API as a whole.

## Nothing here was rewritten

The classes are the classes, moved, with the imports each module actually needs rather than the union every file inherited by sitting in one place. Callers are untouched: `from awx.api import serializers` still works and every `serializers.ThingSerializer` after it still resolves, because the package re-exports each name explicitly, in an order that resolves.

## The cycles, which are the interesting part

Three serializers in `base` name concrete serializers that now live elsewhere: it is how `UnifiedJobSerializer` picks `JobSerializer` or `WorkflowJobSerializer` for a given row. Every one of those twenty references sits inside a method rather than at class level, so the import goes inside the method too and the modules stay acyclic.

The one genuine class-level cycle was `ResourceAccessListElementSerializer`, which subclasses `UserSerializer` and had been sitting with the base classes. It now sits with `user`, where it belongs.

On the views side the only knot was `AdHocCommandList`, which group, host and inventory all build on, so it is extracted first and the other three import it.

## Two things found on the way

- **`views/inventory.py` bound `serializers` to rest_framework's** while every other views module means awx's. It now means awx's here too, and the two `ValidationError` uses import that name directly, which is what the file already does for `ParseError` and `PermissionDenied`.
- **The single serializers module re-exported 123 names it merely imported**, models and helpers and whole modules among them. Nothing outside reaches for those through `serializers.`, with four exceptions, all tests patching a module attribute by string.

## How it is checked

Not by reading the diff. Every URL the API serves was dumped with the view class behind it, and the package asked what it exposes, before and after:

- **418 routes before, 418 after.** Zero added, zero missing, zero whose view class changed.
- **231 view classes** now resolve from a different module, which is the change itself.
- **Every one of the 119 serializer classes is still exported**, and zero changed base classes or `Meta.fields`.

Plus the suites: `awx/main/tests/unit` 1,332 passed and 1 skipped, `awx/main/tests/functional/api` and `unit/api` 1,332 passed, and the whole of `awx/main/tests/functional` run against this change and against the commit before it gives the same 93 failures either way, to the test.

Two tests patched a view class by its old path, and four patched a serializers module attribute by string. All six name where the thing lives now.


cigamit (dismissed):

blaipr:

cigamit (approved):

PR #952, opened 2026-09-12
refactor: make index.html plain HTML instead of a Django template

blaipr:
Sits on top of the stack ending at #951. From the roadmap list in [To Do](https://forum.ascender-automation.org/d/15-to-do-list-of-major-things-on-the-roadmap): "Serve the UI statically: `index.html` is a Django template that staticfiles has to match."

I had this one down as blocked behind retiring styled-components, because of the CSP nonce. It is not. All three nonce uses turn out to be unnecessary.

## `window.NONCE_ID` is read by nothing

```
$ grep -rn "NONCE_ID" awx/
awx/ui/vite.config.mjs:34:    "      window.NONCE_ID = '{{ csp_nonce }}';",
```

Set by an inline script, read nowhere, in any language, anywhere in the tree. It is what is left of `__webpack_nonce__`, which only webpack understood, and webpack went with the Vite move. **That inline script is also the only thing that made `script-src` need to name a nonce at all.**

## The `sc-nonce` meta is belt and braces

It is for the style tags styled-components injects, and the policy already allows those: `style-src 'self' 'unsafe-inline'`. Same for the one inline `<style>`, the app height rule.

## The favicon does not need `{% static %}`

It went through the tag so the build could reach the file through staticfiles. Staticfiles does not hash names here, so the plain path the source file already carries works.

## The result

The built `index.html` has **no Django template syntax in it at all** — the same bytes a static file server would hand out. The vite plugin that injected the template markup now injects only the policy and that one style rule, and is renamed from `djangoTemplate` to `contentSecurityPolicy` for what it actually does.

`installing.html`, the upgrading page, is a real template and still uses the nonce, so the context processor stays.

## Checked in a browser, which is the only check that counts here

Against the development stack, on a page full of styled components:

- **34 style rules injected and applied**, the themed background among them
- **zero content security policy violations**
- no `sc-nonce` meta, `window.NONCE_ID` undefined
- no console errors

Plus the suites: the end to end suite **14 of 14**, and the UI suite **2,989 passing**.

## What is left of the bullet

Serving the file from nginx rather than through Django is now a deployment change in the operator and installer repos, not a platform one. Nothing in the platform requires Django to render it any more.


blaipr:

cigamit (approved):

PR #954, opened 2026-09-13
feat: give the UI a query cache, and put every OPTIONS request on it

blaipr:
Sits on the stack. A roadmap item from [To Do](https://forum.ascender-automation.org/d/15-to-do-list-of-major-things-on-the-roadmap): the UI's reads go through `useRequest`, which caches nothing, so every screen fetches again on every mount.

## What it does

**Gives the UI a query cache.** `@tanstack/react-query`, one `QueryClient` built once and provided at the root, with a thirty second stale time and window-focus refetching off, because a lot of these screens live on a wall display.

**Puts the host facts read on it.** Two screens ask for the same facts, Hosts and the host inside an Inventory, and they now share one request and one cache entry.

**Puts every OPTIONS request on it**, which is the part that reaches the whole application without touching a screen.

## Why OPTIONS is the one worth doing first

OPTIONS describes an endpoint rather than any row in it: what the current user may do with it, and which of its fields are searchable. The UI asks for it **89 times across 174 files**, nearly always inside the same `Promise.all` as the list read beside it. So every list mount pays for it again, and three tabs of the same resource ask three times for one answer.

Caching it at the transport rather than per screen is what makes it free. One line in `defaultHttp` covers all 33 model methods that issue an OPTIONS, and no call site changes shape: `readOptions()` still returns the same promise of the same response. Two components mounting together share one request instead of racing, and a tab switched away and back does not ask again.

The other five methods are left alone deliberately. Four of them write, and a list read is exactly the thing that should not come from a cache unless the screen asking has said so.

Thirty seconds rather than the session, even though an endpoint's shape does not change: the `actions` OPTIONS carries depend on what the user may do, and a role granted while they are looking should not need a reload to show up.

## Two things the implementation has to get right

- **The cache refuses to store `undefined`**, and a transport that resolves to nothing is something a caller may hand us, so the value is boxed. Without that, a caller that used to receive `undefined` would start throwing.
- **The OPTIONS cache gets its own `QueryClient`**, not the one the components use. The two have nothing to say to each other: one is keyed by URL and lives below React, the other is keyed by what a screen asked for and is tied to a mounted tree. Sharing one made the suite flaky, and that is worth spelling out because it is the kind of thing that gets shipped: emptying the shared client between tests cancelled queries a mounted component was waiting on, and a different file failed each run. With them separated the suite runs identically to the commit before this one, three times over, where before it differed in two runs out of four.

## Checks

Five tests cover the cache directly: two callers at once cost one request, a later caller is served from the cache, two endpoints do not answer for each other, reads and writes stay uncached, and a failure is not cached so the next caller tries again.

The UI suite runs to the same 17 files and 24 failures as the commit before this one, which are the pre-existing timezone assertions, three runs in a row.

## What this does not do

`useRequest` still has 304 call sites, 185 of them reads. Those are not a mechanical swap: each one decides what its query key is and when it should be invalidated, so each is a behaviour change that wants reviewing on its own screen rather than in a batch of a hundred and seventy. The cache and the pattern are here now, and the list screens are the obvious next ones, since that is where a refetch on every mount actually costs something.

Also fixes `useHostFacts.test.tsx`, which imported `ResponseOf` from a `testUtils` module that does not exist. vitest resolved it and tsc did not.


blaipr:
The rest of the list screens now read through the cache: **thirty five of the thirty nine**.

What the first pass got wrong about the remainder was smaller than it looked. Some name the request before the result rather than after, some read `search` off `useLocation` rather than `location.search`, and some build `QS_CONFIG` from a variable rather than a literal. None of those is a behaviour difference, and none of them justified leaving the screen behind.

**Four are left alone, with reasons rather than excuses:** two call `setPendingHealthCheck` inside the read, and a read that also sets state is not a read, so caching it moves when that fires; one has no plain mount effect to drop; one builds its read differently enough that the pattern does not match.

## Two real bugs came out of finishing it, both about the key

**Keying on the `QS_CONFIG` namespace put two different screens on one entry, twice.** The host list and an inventory group's host list both said `'host'`; the team list and an organization's team list both said `'team'`. Different endpoints, one cache entry. Every key is now the screen's own name.

**Worse, nine of these reads take a route parameter and the key did not.** A user's tokens and the next user's tokens are the same URL search with a different id in the path, so the second would have shown the first's rows until the entry went stale:

```
['user-token-list', id, location.search]
['inventory-group-host-list', groupId, location.search]
['inventory-related-group-list', groupId, inventoryType, location.search]
```

Checked after: **35 keys, 0 of them shared.**

Neither bug was visible in the first commit, because one screen on a cache cannot collide with itself. They only appear once enough screens are on it, which is an argument for doing the rest rather than stopping at fifteen.

The UI suite is **553 files, 3,005 tests, 0 failures**, with tsc, eslint and prettier clean.

blaipr:

cigamit:
NVM, bad merge in my environment.

cigamit (approved):

PR #963, opened 2026-09-13
refactor: take the UI off formik

blaipr:
formik is out of the UI: out of `package.json`, out of `node_modules`, and its licence file with it. 681 lines of field layer stand where it did, and the 183 files that used it changed an import rather than a signature.

## The order, which was the point

1. **The contract first**, written against formik so it described what was there rather than what was wanted.
2. **The layer**, held to that contract and nothing else.
3. **The field components**, which could move while every form above them was still a formik one, because `useField` read both contexts and used whichever answered.
4. **The 191 form roots.**
5. **The bridge out.**

The risk in leaving formik was never the API. It was the behaviour the API implies and nothing states: when a validator runs, when a field counts as touched, what `isValid` says about a pristine form whose required field is empty. So that is what the contract records, and for as long as the tree held both implementations every test ran twice, once against each, with a difference in either direction a failure.

## Three things the measurement got wrong

Each was found by running the suite rather than by reading it, and each is worth more than the code it cost.

- **Field names are not all flat.** The first count looked for quoted literals and missed the ones built as template literals: `${prefix}.startDate` in the schedule subform, `inputs.${id}` in the credential fields. A name is a path, and reading, writing, touching and validating one all walk it.
- **A change event does not just carry `target.value`.** A number input hands over the string `"134"`, and `OrganizationForm` caught it, sending `max_hosts` as `0`. Numbers, ranges, checkboxes resolved against what the field holds, and multiple selects each mean something different.
- **The form root takes more than `initialValues` and `onSubmit`.** The regex that said otherwise stopped at the first `>` and never saw a multi-line tag. `ScheduleForm` passes a form-level `validate`, two forms pass `enableReinitialize`, and `UserForm` reads `setErrors` off the bag `onSubmit` is handed as its second argument.

## And one bug of mine, which React caught rather than a test

Validation read the current values by calling `setValuesState` with a function that also set errors, which is a `setState` during render. The Login form reported it as updating one component while rendering another, and it was right. Values are held in a ref now and nothing nests.

## What the layer is

A form root taking `initialValues` and `onSubmit`, a `useField` returning the triple every call site already destructured, a context with the members the tree reads, and `withForm` for the three large forms that were built with `withFormik`. Validators are the plain functions `util/validators` already exported. Validation runs every registered validator on every change, which is what formik does and what makes an error clear at the same moment as the one that replaced it.

`FormRoot` rather than `Form` at the call sites, because 47 of the files that hold one already bind `Form` to PatternFly's, which is the markup this is the state behind rather than a competing version of it.

## The ratchet is gone

It existed to stop the coupling growing before an exit existed. Counting screens that import formik is not a measurement any more.

The contract stays, at 34 behaviours, and the table it is built on is left in place so a second implementation can be put beside this one again if there is ever a reason.

## Checks

- The UI suite fails on the same **17 files** as the commit before this, which are the pre-existing timezone assertions.
- The **14 end-to-end tests** pass against a build with no formik in it.
- `node_modules/formik` does not exist.


cigamit (dismissed):

blaipr:

cigamit (approved):

PR #968, opened 2026-09-13
test: pin the suite's timezone, so it says the same thing everywhere

blaipr:
The UI suite asserts on formatted local times. `formatDateString('2018-01-31T01:14:52.969227Z')` is expected to be `1/31/2018, 1:14:52 AM`, which is only true in UTC, and 23 assertions across 17 files are like it.

GitHub's runners are UTC, so CI has always been green and nothing said the dependency was there. On a developer's machine anywhere else the suite opens with two dozen failures that have nothing to do with their change, and the first thing they have to work out is which of the failures are theirs.

## What it does

One setting in `vitest.config.mjs`:

```js
env: {
  TZ: 'UTC',
},
```

No assertion changes. The suite keeps asserting on local time, and local time is now the same everywhere it runs.

## Checked

From a CEST machine, which is where the problem shows:

- **Before**: 24 failures across 17 files, every one a formatted date or time.
- **After**: none. 2,991 tests across 551 files, all passing.

The same run with `TZ=UTC` set by hand in the shell gave the same result before this change, which is what identified the cause rather than guessing at it.


cigamit:
Seems these changes were already merged via another PR

blaipr:
Correct, and this branch is now empty: `git diff origin/main` against it returns nothing.

The setting landed in #954, which carried `TZ: 'UTC'` in `vitest.config.mjs` as part of that stack. Closing.

PR #972, opened 2026-09-13
perf: stream the job stdout download instead of buffering it

blaipr:
This started as the measurement behind the async views item, and turned up something else on the way.

## What I set out to check

The roadmap and [Future of Ascender](https://forum.ascender-automation.org/d/2-future-of-ascender/9) both say an async request path would stop streaming job output holding one of five uwsgi workers per viewer. That is the payoff for a rewrite, so it wanted a number.

It is wrong, and the routing says so before any measurement does: `nginx.locations.conf` sends `/websocket/` to daphne and everything else to uwsgi, and `JobOutput.tsx` subscribes through `connectJobSocket`. Live output never reaches a uwsgi worker.

Measured against the development stack anyway, eight viewers on a running job:

| | |
| - | - |
| websocket frames in 20 seconds | **1,376** |
| HTTP requests those viewers made | **0** |
| busy uwsgi workers, sampled once a second | **0 of 5** |

There is no `StreamingHttpResponse` anywhere in `awx/api/` or `awx/main/` either. So that bullet buys nothing, and I will correct the post rather than leave it standing.

## What does hold a worker

The stdout download, and it holds one for a reason worth fixing.

`result_stdout_raw_handle()` takes care not to put a download on the heap. It `COPY`s the event stdout straight to a temporary file instead of going through the ORM, and says why in a comment: *"we do not want to generate queries here that construct model objects by fetching large gobs of data (and potentially ballooning memory usage)"*.

The view then handed that file to `HttpResponse`. Its content setter does `b"".join(...)` over the iterator on assignment, so every byte the model kept off the heap goes straight back on it.

## The change

One class. `StreamingHttpResponse` is the one that respects an iterator, and `FileWrapper` already hands it 8 KiB at a time.

Measured on a job whose download is 11.6 MB, sampling every uwsgi process from `/proc`:

- **one download**: a worker's resident memory grew **45 MB**, now **21 MB**.
- **five at once**: **146 MB** across the pool, now **51 MB**.
- **nine sequential downloads**: **404 ms** median before, **392 ms** after.
- **the bytes**: identical, 12,155,254 either way.

Nothing gets faster. The pool stops paying for the file twice.

## The trade

`Content-Length` goes, because the filters rewrite the content as it streams and no length is known before the last line is read. The response is chunked, so a browser downloading a large job's output shows no total size. `Content-Disposition` is untouched, so the file still saves under its own name.

That is the one user-visible cost, and worth naming rather than burying.

## Why this is not an argument for async views

It is a synchronous read of a file on disk. Async would not have helped it, and the fix is the same either way. The async bullet still has `ATOMIC_REQUESTS` behind it and the middleware work in #964, but it no longer has this.

## Checks

- `test_download_is_streamed_in_blocks` covers all four job types and both download formats, asserts more than one block comes out of the iterator and none exceeds 8 KiB, and **fails on the commit before this one**.
- `awx/main/tests/functional/api/test_unified_jobs_stdout.py`: 67 passed.
- `awx/main/tests/functional/api/`: the same **8** failures before and after.
- The swagger sample recorder in `conftest.py` now skips streaming bodies, since draining that iterator would hand the test an empty response.


blaipr:
Up to date with main and mergeable.

cigamit (approved):

PR #974, opened 2026-09-13
refactor: say which object a read follows instead of writing the query

blaipr:
Sits on top of #950. A second slice of the roadmap item in [To Do](https://forum.ascender-automation.org/d/15-to-do-list-of-major-things-on-the-roadmap): "Declarative permissions instead of 3,103 lines of per-model access classes, one place to change."

#950 took the safest property, read only, and said it once. This takes the largest one.

## Whether "rules as data" is even the right framing

I went looking before writing anything, because #950 found only nine of fifty classes shared a property and that reads like the idea not fitting. Classifying all 145 write rules by shape:

| | | |
| - | - | - |
| constant, including superuser only | **25** | 55 lines |
| delegation to another access class | **4** | 9 lines |
| a single expression | **35** | 91 lines |
| branching, raising, inspecting the payload | **81** | 721 lines |

So on the write side the answer is no, and not marginally: 82% of the lines are imperative, and of those, 55% branch on the proposed new state rather than on the user and the object, 16% raise a specific message instead of returning a boolean, and 16% consult the licence. A permissions table maps subject, object and action to a verdict. These also take the payload and answer with a reason. **That is a different shape, and the roadmap bullet as written overstates what is available.**

The read side is the opposite. Of 42 `filtered_queryset` definitions the median body is **one line**, and twenty of them are the same sentence: show the rows reachable from an object this user may read. They differ only in which model carries the `read_role` and the ORM path to it.

## What this does

Nineteen of those twenty now name the two things instead of writing the query:

```python
class HostAccess(BaseAccess):
    model = Host
    read_via = (Inventory, 'inventory')
```

`ReceptorAddressAccess` is the twentieth and is deliberately left out, because its rule is not this one. See #973.

## Why it is worth doing, which is not the 38 lines

`ReceptorAddressAccess` is the whole argument. It wrote its filter by hand against `Instance`, which is not a `ResourceMixin` and has no `accessible_pk_qs`, so it raised `AttributeError` for every user `get_queryset()` did not short circuit: everyone who is neither superuser nor system auditor. It has been that way since the 23.8.0 upstream sync in February 2024, and nothing caught it, because there were nineteen other hand-written copies and no shared place to check.

`TestReadVia` asks every declaration whether its model carries the role and whether Django can follow its path. Declaring the broken receptor address rule fails it:

```
AssertionError: ReceptorAddressAccess: Instance has no role accessor
```

## Nothing changes, and it is checked rather than argued

Every access class in the registry was asked for its read queryset as a plain user, an org admin, a system auditor and a superuser, and the SQL compared either side of the change:

- **196 signatures, 192 byte-identical.**
- **4 differ**, all `WorkflowJobTemplateNodeAccess`, which passed `accessible_objects` into an `__in` where the other eleven paths pass `accessible_pk_qs`. The shared path uses the latter, which drops one subquery level through `main_workflowjobtemplate`.

For those four, the row sets were compared directly instead: three users, four templates, several different roles, and then a template deleted with its roles left behind, which is the one case where the extra subquery level could have mattered. **6 comparisons, 6 identical.**

## One thing removed on the way

`InventoryAccess.filtered_queryset` took `allowed` and `ad_hoc` arguments that nothing in the tree passes and the body never read.

## Checks

- `awx/main/tests/unit/test_access.py`: 42 passed.
- `awx/main/tests/unit` and `awx/main/tests/functional` matching rbac, access, activity, inventory, workflow or credential: the same **43** failures before and after.


blaipr:
Folded into #950, which is the pull request for this roadmap item. Same commit, same measurements.

blaipr:
Reopened and held for reuse. The commit on this branch is also on #950, which is the pull request for this roadmap item, so this one should not be merged as it stands: take the read access change from #950.

blaipr:
This branch is now empty: `git diff origin/main...` against it returns nothing.

The work landed in #950, which carries the declarative `read_via` form, 23 of them in `awx/main/access.py` on main. Closing.

PR #975, opened 2026-09-13
feat: serve the API and the websockets from one uvicorn

blaipr:
Reusing this number rather than opening another: it was already the ASGI pull request, and its contents duplicated #942, which is the thing this split resolves.

**#942 now carries the fix, and this carries the trade.** They were one pull request, and only one of them is a judgement call.

## What this does

Replaces uwsgi and daphne with a single uvicorn. One program where there were two, and uwsgi comes out of the requirements.

Sits on #942, which is what makes it possible: the ASGI application could not answer HTTP until that landed.

## Verified with both old servers stopped

Development stack, nothing left but nginx and uvicorn:

- the API answers, `/api/v2/me/`, `config`, `organizations` and `settings/ui` among them.
- job output **still streams over the websocket**: 94 frames in twelve seconds against 96 for daphne, same 81 rows rendered.
- the one console message, a websocket closing before it is established, appears against daphne too, so it is not new.

## The cost, which is the whole decision

Through nginx, 60 requests at each concurrency, three runs:

| | uwsgi + daphne | uvicorn |
| - | - | - |
| `/api/v2/ping/`, 5 concurrent | **41.1 req/s** | 14 to 18 |
| `/api/v2/ping/`, 20 concurrent | **45.2 req/s** | 24 to 31 |
| `/api/v2/job_templates/`, 5 concurrent | **4.6 req/s** | 2.7 to 3.1 |
| `/api/v2/job_templates/`, 20 concurrent | 4.6 req/s | **4.2 to 5.0** |
| one at a time | 78ms / 497ms | 86ms / 511ms |

The heavier endpoint is at parity under load. The trivial one is about half.

## And there is no payback coming

I originally argued this cost was what the async views item would remove, so the two wanted deciding together. **I measured that and it is wrong.** Two views doing the same twenty row ORM read, one `def` and one `async def`, under uvicorn: async is about a fifth slower one request at a time and level under load, and Django's own async ORM API is no better. The ORM is synchronous, so an async view still crosses into a thread to read the database and the hop moves rather than goes. Django refuses the pair outright anyway, with `RuntimeError: You cannot use ATOMIC_REQUESTS with async views`.

So this is the final cost, not an interim one.

## What I would want before merging it

My numbers come from one development container, five workers, TLS through nginx, on seeded data. That is enough to say uvicorn is viable and daphne is not, and thin for a change to how every deployment serves traffic.

Two things would settle it: **the same benchmark on a production shaped deployment**, and **a look at what actually polls the light endpoints**. If that is a load balancer health check at a sane interval, one server instead of two and one less C dependency is clearly worth it. If something hammers `/api/v2/ping/`, it is not.

That is why it is here on its own rather than holding up #942.

## Two things deliberately kept

**daphne stays as a dependency.** It is the Django app that gives `runserver` its ASGI handler, and `defaults.py` lists it in `INSTALLED_APPS` for that. What goes is daphne as a *server process*.

**The uwsgi branch in `statement_timeout.py` stays**, for a deployment still running uwsgi out of tree, with its comment corrected to say uwsgi is no longer what we ship.

## One note on the lockfile

`requirements.txt` was edited to add uvicorn's closure and drop uwsgi, rather than recompiled. A full `uv pip compile` carried **179 lines** of unrelated version bumps, and this repository keeps those to one pull request each so a bad one stays bisectable.


blaipr:
Folded into #942, which is the pull request for this roadmap item. Same commit, same measurements.

blaipr:
Reopened and held for reuse. The commit on this branch is also on #942, which is the pull request for this roadmap item, so this one should not be merged as it stands: take the ASGI change from #942.

blaipr:
All three applied. Each was something uwsgi was doing that nothing picked up.

**The mount path.** Right, and the 404 was real: the compose stack still exported `UWSGI_MOUNT_PATH`, uwsgi consumed it as a mount, and uvicorn reads no such thing. It is `AWX_MOUNT_PATH` now, and the `uvicorn` target passes it as `--root-path`, which is the same instruction to an ASGI server. Unset, which is the default, the flag is not passed at all:

```
AWX_MOUNT_PATH=""        -> uvicorn --host 127.0.0.1 --port 8051 awx.asgi:channel_layer
AWX_MOUNT_PATH="/prefix" -> uvicorn --host 127.0.0.1 --port 8051 --root-path "/prefix" awx.asgi:channel_layer
```

**The development marker.** Also right, and worth spelling out why it was invisible: `is_web_process()` answered by importing uwsgi, so the dev supervisor never needed to say anything. With uwsgi gone the import fails and the answer flips. `environment = AWX_WEB_PROCESS=1` is set there now, as production sets it. Checked against the real settings module:

```
AWX_WEB_PROCESS=''  -> options: none
AWX_WEB_PROCESS='1' -> options: -c statement_timeout=110000
```

**The request deadline.** You are right that there is no equivalent, and I would rather the template say so than imply the protection survived.

The dominant case is covered. harakiri mostly caught a query that would not come back, and a web process gets `DATABASE_STATEMENT_TIMEOUT`, so PostgreSQL cancels at 110 seconds where harakiri killed the worker at 115.

What is not covered is exactly what you describe: a handler blocked without touching the database. `--timeout-worker-healthcheck` does not reach it, because the healthcheck is answered by the event loop and a synchronous view runs in a thread, so the worker keeps looking healthy. There is no per-request deadline in uvicorn to set. `--limit-concurrency 512` bounds how far it spreads, refusing with 503 rather than queueing without limit, and the comment above the program records both halves so the next person does not go looking for a harakiri that is not there.

cigamit (approved):

PR #976, opened 2026-09-13
refactor: give the settings categories one edit form

blaipr:
Sits on top of the stack ending at #963.

Twenty one settings categories each had their own edit screen, and each did the same work around its own list of fields: read the category, merge every value with the PUT options describing it, build initial values, save, revert, cancel, and render the loading, error, form, revert button and confirmation modal scaffolding.

The only things that actually differed were **the category name, where to navigate back to, and which fields to show**. `SettingsEditForm` takes the first two and a render prop for the third.

## What a screen becomes

```tsx
function RADIUSEdit() {
  return (
    <SettingsEditForm category="radius" detailUrl="/settings/radius/details">
      {(radius) => (
        <>
          <InputField name="RADIUS_SERVER" config={radius.RADIUS_SERVER} />
          <InputField name="RADIUS_PORT" config={radius.RADIUS_PORT} />
          <EncryptedField name="RADIUS_SECRET" config={radius.RADIUS_SECRET} />
        </>
      )}
    </SettingsEditForm>
  );
}
```

That is the whole file, where it was 131 lines.

| | |
| - | - |
| screens moved | **15** of 21 |
| lines before | **2,649** |
| lines after | **755**, plus **204** in the shared form |

## The six that stay, and why

The rewrite checks each screen against what the shared form actually does rather than assuming, and says why it declines:

- **three** merge the options their own way: MiscAuthentication, MiscSystem, UI.
- **one** navigates to more than one place: LDAP.
- **one** builds initial values differently: Logging, whose fields also read the form state directly.
- **one** does not read a plain category: Subscription.

Jobs and Troubleshooting were in this list and are not any more: both turned out to want the `except` prop and nothing else, so they moved.

## Three things I got wrong first, each caught by the suite

Worth writing down, because each one is a behaviour difference hiding inside what looks like the same code.

**The options guard.** Seventeen of the twenty one skip a value the PUT options do not describe; four keep it. Merging everything renders a field with no label, no help text and no type, and the CodeMirror editor inside it throws `(config.doc || "").split is not a function`. 178 errors, from one missing `if`.

**Initial values are not a subset.** Fourteen turn a list or a nested object into JSON, because the editor takes text. The plain variant looks like a subset of that and is not: a screen using it beside an object field submits the value as itself where the shared form submits JSON, and two tests said `expected 'string' to deeply equal 'object'`. The rewrite now takes a plain one only when the screen has no list or object field at all, which is how RADIUS and TACACS still qualify.

**Submitting has to undo the stringifying.** Screens named each object field and ran `formatJson` over it by hand. The configuration already says which fields are lists and nested objects, so the shared form asks it rather than being told, which is the mirror of what `initialValues` does going in.

## Checks

- The UI suite: **551 files, 3,025 tests, 0 failures.**
- `tsc --noEmit` clean, `eslint` clean.
- `prettier --check` reports the same three pre-existing files before and after.

Every one of these screens already had a test covering render, submit, cancel, revert and the error path, and all of them pass unchanged. That is the check that matters here: nothing in the tests was adjusted to fit the refactor.

## One commit that is not mine

The first commit takes the timezone pin from #968, so this branch runs clean on a machine that is not UTC. Whichever of the two lands first, the other drops it.


blaipr:
Two more screens on the shared form, so **fifteen of the twenty one**.

Jobs and Troubleshooting were left alone in the first pass because the rewrite could not tell what they did differently. What they needed turned out to be one idea stated two ways: a category can return more than its screen shows, and screens said so by hand. Jobs destructures two stdout display limits out before merging; Miscellaneous System plucks the nine it wants. The shared form now takes `only` or `except` and does it itself.

Troubleshooting needed nothing at all. Its `const { ...debugData } = data` is a copy, so it was the standard read wearing a different variable name, and the first pass declined it for exactly that.

| | before | after |
| - | - | - |
| `JobsEdit.tsx` | 276 | **119** |
| `TroubleshootingEdit.tsx` | 152 | **32** |

Two things in Jobs worth pointing at rather than leaving in the diff:

- Its fields asked `options?.X` for whether a setting is required. The merged configuration only holds keys the options describe, so `jobs.X` answers the same question about the same key, and it is the object the field already takes.
- It relabels one choice, because the API calls the template option "Only Template Variables and Definitions" where the help text and the detail screen both say template. That derivation moved into the render prop, which is where per screen work belongs now.

**Six screens keep their own form, each for a reason:**

- **Miscellaneous System** reads an execution environment and maps it to an id on submit.
- **Miscellaneous Authentication** reads a second category and merges two option sets.
- **Subscription** is a wizard rather than a category form.
- **LDAP** groups its settings into a nested shape.
- **Logging** builds initial values differently and its fields read the form state directly.
- **UI** navigates with state and mutates the options object it was handed.

Those are not "not done yet": pushing any of them through a shared form would mean either a prop that exists for one caller, or changing what the screen does. Worth doing separately if at all.

The UI suite is **551 files, 3,025 tests, 0 failures**, with tsc, eslint and prettier clean.

blaipr:
Both looked at. One is a real error in the description and is fixed; the other I went and checked, and I do not think it changes what gets saved.

## The scope was wrong, and is corrected

You are right. **15 of 21**, not 13, and Jobs and Troubleshooting are among them rather than exceptions. They were in the "stays" list when that was written and both turned out to want the `except` prop and nothing else, so they moved and the summary did not follow. The body now says 15, lists the **six** that stay (LDAP, Logging, MiscAuthentication, MiscSystem, Subscription, UI), and the line counts are recomputed: 2,649 before, 755 after plus 204 in the shared form.

## The guard: real difference, no effect on what is written

The behaviour difference is real, and so is the trigger. Of the seventeen screens this touches, **fifteen had the guard on main and RADIUS and TACACS did not**, so for those two a key present in `GET` but absent from the `PUT` options used to survive into the form and go back out on save.

And a setting defined in a settings file is exactly such a key. `awx/conf/settings.py` marks it at runtime:

```python
if file_default != init_default and file_default is not None:
    logger.debug('Setting %s has been marked read-only!', key)
    self.registry._registry[key]['read_only'] = True
    self.registry._registry[key]['defined_in_file'] = True
```

So it becomes `read_only`, and a read-only field is absent from the `PUT` actions, which is why it fell through the guard.

What it cannot do is change what is saved. That field is built with `read_only=True`, and DRF drops read-only fields from input, so submitting the value was already a no-op. The old behaviour was sending something the serializer discarded.

What the guard does change is that the field stops rendering. That is the point of it: with no entry in the `PUT` options there is no label, no help text and no type to render from, which on other screens produced 178 CodeMirror errors before the guard existed. SAML is the case that shows it working as intended, since it has two genuinely read-only settings, `SOCIAL_AUTH_SAML_CALLBACK_URL` and `SOCIAL_AUTH_SAML_METADATA_URL`, and it already had the guard.

So rather than keeping a no-guard path for two screens, this makes RADIUS and TACACS behave as the other fifteen already did. Happy to add the prop if you would rather have the choice available, but I did not want to carry a branch whose only effect is to submit values the API throws away.

cigamit (approved):

PR #979, opened 2026-09-13
bug: Workflow Always link duplicated the Success color

cigamit:
The Workflow "Always" link on the default theme had the same color as the Success.  This changes it to give it its own css color variable, and we clean things up a bit while we are at it.


cigamit:
I'll redo this after all the changes

PR #987, opened 2026-09-14
fix: spool external logs somewhere writable when the setting is not

blaipr:
`LOG_AGGREGATOR_MAX_DISK_USAGE_PATH` is an API setting, so anyone with access to Settings can point it at a directory rsyslog cannot write. The code knows that and falls back, except the fallback sat **below** the line that writes `queue.spoolDirectory` into the config:

```python
    queue_options = [
        f'queue.spoolDirectory="{spool_directory}"',
        ...
    ]

    if not os.access(spool_directory, os.W_OK):
        spool_directory = '/var/lib/awx'      # nothing reads it again
```

`spool_directory` is never read after that line. The generated config named the unwritable directory anyway.

## What that costs

rsyslog has nowhere to spool, so the disk queue stops surviving a restart, which is the one thing the thirteen queue options exist for. Nothing reports it: the config is valid, rsyslog starts, and external logs are simply lost on the next restart instead of being delivered late.

It has been this way since the initial import, so it is inherited from upstream AWX rather than something the fork introduced.

## The fix

Moving the check above `queue_options`. The fallback also logs now, because silently writing somewhere other than the configured path is how this went unnoticed.

## Checking

Three tests: the configured directory is used when it is writable, the fallback is what reaches the config when it is not, and the fallback says so. The two that describe the bug fail against the old code; the first passes either way and is there as a regression guard.

1,532 unit and conf tests pass. `ruff check` and `ruff format --check` are clean. The one red test on this branch, `test_every_setting_this_process_can_read_is_described`, is already red on `main` and is what #984 fixes.

Touches lines next to #985, so whichever lands second needs a trivial rebase.


cigamit:
`/var/lib/awx`
Didn't another PR just change that to `/var/lib/ascender`?

blaipr:
Both addressed.

**On `/var/lib/awx`:** yes, [#985](https://github.com/ctrliq/ascender/pull/985) moves it to `/var/lib/ascender`. This PR is cut from `main` so it still says the old name, and the two touch adjacent lines. It is now a single module constant, `DEFAULT_SPOOL_DIRECTORY`, rather than the same literal twice, so whichever of the two lands second is a one-line rebase.

**On the permission check:** right, and the old check was weaker than it looked. `os.access(path, os.W_OK)` says yes to a plain file, and to a directory with the write bit but not the execute bit, and rsyslog can spool into neither: it creates its queue files *inside* that path. It is `os.path.isdir` plus `W_OK | X_OK` now, with a test for each of the two cases the old check let through.

cigamit (approved):

PR #989, opened 2026-09-14
test: pin the job folder prefix two other things depend on

blaipr:
`JOB_FOLDER_PREFIX` names the directory a job's private data goes in, and two things outside `awx/main/constants.py` read the exact string.

- `Instance.get_cleanup_task_kwargs` builds the glob that sweeps stale job folders from it. A folder written under one prefix and swept under another is never cleaned up.
- **ascender-kit** reads it, in `ascenderkit.api.pages.unified_jobs`, to find the host side of a job's volume mount. A mismatch means the lookup silently returns nothing.

Neither breaks at build time, which is why the constant carried this:

```python
# also update awxkit.api.pages.unified_jobs in ctrliq/ascender-kit if changed
```

Two things wrong with that. It names `awxkit`, which has been `ascenderkit` since #660, so the next person greps for a module that does not exist. And it is a request rather than a check.

## What the tests say

- The exact string ascender-kit builds `/tmp/awx_<id>` from.
- The trailing separator. Without it, cleaning up job 4 would match job 42's folder, because `awx_4_` would be a prefix of `awx_42_`.
- The single `%s`, since one caller does `% '*'` to build the cleanup glob and the other does `% instance.pk` to name a folder.

The value itself does not change. `awx_%s_` is a path on disk that a running deployment already has folders under, and moving it is its own decision, with the cleanup sweeping both prefixes.


cigamit:
Seems it brought in a lot of stuff from the build directory that should have gitignored.

blaipr:
Fixed, it is two files now: `awx/main/constants.py` and the new test.

You were right about the cause. A local branch had moved `awx/` to `ascender/`, which took the build output with it, and `.gitignore` names the build directories under `awx/` rather than by what they are. So on a branch cut from `main` the moved copies were neither tracked nor ignored, and `git add -A` swept them up.

cigamit (approved):

PR #994, opened 2026-09-14
Fix root path making app unaccessible

cigamit:
`django.request Not Found: //api/`

PR #975 replaced uwsgi and daphne with a single uvicorn. 

The `make uvicorn` recipe passes `--root-path "$AWX_MOUNT_PATH"` whenever that variable is set, and the docker-compose template always sets it from the `ingress_path` default of `/`. 

uvicorn 0.52 builds the request path as root path plus path, so `/` plus `/api/` became `//api/` and Django 404'd everything. uwsgi's old `mount` directive treated `/` as "no prefix", which is why this never showed before. 

nginx was not involved: its `proxy_pass` has no URI so it forwards the path verbatim, and the nginx access log confirms the browser sent plain `/api/`.


cigamit:
Clsoing, moving to #996 as it fixes a CI test

PR #997, opened 2026-09-14
refactor: rename the awx package to ascender

blaipr:
```python
>>> import ascender.main.models, awx.main.models
>>> awx.main.models is ascender.main.models
True
```

The package on disk is `ascender/` now, and `awx/` is the alias. Same machinery as #988, the other way round: every module answers to both names and resolves to one module object.

## The old name cannot be dropped, and not only for a release

The `awx.credential_plugins` entry points live in other people's packages and cannot be rewritten from here.

The dispatcher is the one that cannot wait for anything to move, and it is where this nearly went wrong. `TaskWorker.resolve_callable` refused any task name not starting with `awx.`:

```python
if not task.startswith('awx.'):
    raise ValueError('{} is not a valid awx task'.format(task))
```

After the rename every task name starts with `ascender.`, so **the dispatcher would have rejected every task it published**. It takes both prefixes now, which is also what a message queued before an upgrade needs in order to resolve after it. Four tests cover that, and nine tests fail without the change.

## What moved

2,035 import statements, the module paths in `INSTALLED_APPS`, the app configs, the settings module, the middleware and filter class paths, the periodic task paths, the mock targets in the tests, and the packaging, Makefile, Dockerfile, tox, CI and documentation paths.

## What deliberately did not move

Each of these is a published name rather than an import path, and two of them broke before I worked out why:

- **The logger names.** All 155 are identical to `main`, checked by extracting them from both trees and comparing. The five loggers that derived their name from `__name__` are now named explicitly, or they would have slipped out from under the `LOGGING` keys that configure them.
- **`LOGGER_BLOCKLIST`**, a tuple of those logger names. Moving it stopped `awx.conf` records being blocked, so the level filter read `LOG_AGGREGATOR_LEVEL` for each one and preloaded the settings cache from the database. The suite went from seventeen seconds to fourteen minutes.
- **`CREDENTIAL_PLUGIN_GROUPS`**, where `awx.credential_plugins` is the group a third party declares itself under. Moving it left both entries naming the same group, so nothing outside this repository would have loaded at all.
- **The distribution name** in `pyproject.toml`, still `awx`, so `dist/awx.tar.gz` and every installer that names it are untouched.
- The venv at `/var/lib/awx/venv`, the image layout, and the container names.

## Django app labels are unaffected

A label is the last component of the app config's name, so `awx.main` and `ascender.main` are both `main`. No migration changes, no content type changes.

## Checking

The full suite CI runs, unit, functional, conf and sso: **4,208 passed, 2 failed**. The same scope on `main` gives **4,203 passed, the same 2 failed**, so neither is from this change:

- `test_python_and_js_licenses`, which #993 has since fixed and this branch now carries.
- `test_named_url_mappings_exists_only_once_the_middleware_is_built`, which passes in isolation and fails in the combined run because `NAMED_URL_FORMATS` has left the settings registry by the time it runs. Order dependent, and identical on `main`.

I ran the baseline rather than arguing from the code, because two of the three bugs above looked like they could not possibly be related to a rename until they were. `ruff check`, `ruff format --check` clean over both packages. `ascender-manage check` runs. Django boots under `ascender.settings.development` and under `awx.settings.development`, with `ROOT_URLCONF`, `WSGI_APPLICATION`, `ASGI_APPLICATION`, `INSTALLED_APPS` and the periodic task paths all resolving, and one `main` entry in the app registry.

Eight tests cover the alias itself, each in its own interpreter, since imports are cached for the life of a process and a warm one proves nothing about a cold start.

## Merge order

This touches every file under `awx/`, so it should go last of the open Ascender PRs; anything merged after it needs a rebase onto the new paths.


blaipr:
Rebased, conflicts resolved, mergeable.

cigamit:
The only thing I notice, and its probably not a blocker, but this shows up in the logs multiple times when bringing up the container.

```
System check identified some issues:
WARNINGS:
?: (staticfiles.W004) The directory '/awx_devel/ascender/ui/build/static' in the STATICFILES_DIRS setting does not exist.
```

blaipr:
Good catch, and it is this PR's doing. Fixing it, pushed.

`bootstrap_development.sh` creates that directory on purpose, with the comment "Make sure that the UI static file directory exists, Django complains otherwise". The rename never touched the file, so it still ran `mkdir -p /awx_devel/awx/ui/build/static` while the package had moved to `ascender/`. The directory it created was no longer the one `STATICFILES_DIRS` points at, so the warning came back.

Looking for the rest of that pattern found four more places the rename left behind, all of them pointing at a path that no longer exists:

- `awx-autoreload` was watching `/awx_devel/awx`, from the Makefile and from all three supervisor templates, so nothing was reloading on a source change.
- Its own exclude list still read `/awx_devel/awx/ui`, so once the watch path is corrected it would restart on every UI and test file too.
- The dev nginx favicon alias still pointed at `/awx_devel/awx/public/static/favicon.ico`.
- `ui-api-types` had `awx/ui/node_modules` as a prerequisite, which is the one CI caught.

The `/var/lib/awx` paths are deliberately left alone here, since moving the data directory is its own step.

blaipr:
Still working on this one, quite a bit has gone on top since you looked. I will give you a shout @cigamit  when it is ready for review rather than have you read a moving target. Thanks!

blaipr:
@cigamit this one is ready for a run of the suite whenever you have a moment.

The three jobs that failed on the last run that executed are fixed:

- `api-lint` and `ui-lint`: `ascender-` is longer than `awx-` and `tower-`, which pushed five Python files and fifteen TSX files past the line limit. Reformatted with the pinned ruff and prettier.
- `ui-test-general`: the About modal asserts the brand in the speech bubble, and the default moved from `AWX` to `Ascender Automation`. The assertion follows it.

Locally the UI suites are green, 1118 general and 368 screen files, and the Python unit suite passes.

cigamit:
Looks like extract-strings just needs to run again

blaipr:
Both fixed, thanks.

`extract-strings` was the whole of the `ui-lint` failure, and running it alone would have been the wrong fix. The string that moved is the provisioning callbacks help text, `Ansible AWX` to `Ascender`. A changed msgid is a new entry to lingui, so the eight translations of that sentence came back empty and the screen would have shown English to every non-English reader. Each one is carried onto the new id with the brand substituted, which is the only word in the sentence that moved.

The `staticfiles.W004` is the ordering. `bootstrap_development.sh` created `ascender/ui/build/static` after the migration step, and the system check warns once per management command while it is missing, which is why it appears several times rather than once. The `mkdir` moves to the top, ahead of `make ascender-link` and `make migrate`. Running `ascender-manage check` against a tree with the directory absent and then present gives one issue, then none.

Locally `ui-lint` now passes all three of `lint`, `prettier-check` and `check-strings`, and both UI suites are green.

cigamit (approved):

PR #72, opened 2025-04-09
Add option for SAMESITE

cigamit:
Resolves #48

Patch from Upstream
https://github.com/ansible/awx/pull/15100

PR #998, opened 2026-09-14
refactor: give the settings #984 could not reach the Ascender name

blaipr:
#984 renamed the eleven settings the database registers. **Seventeen more were left**, because they are not registered: thirteen are defined in `defaults.py` and overridden in a settings file, and four are read from the environment.

## The thirteen

`ASCENDER_AUTO_DEPROVISION_INSTANCES`, `ASCENDER_CALLBACK_PROFILE`, the five `ASCENDER_CONTAINER_GROUP_*`, `ASCENDER_CONTROL_NODE_TASK_IMPACT`, `ASCENDER_NOTIFICATION_REQUEST_TIMEOUT`, `ASCENDER_REBUILD_SMART_MEMBERSHIP`, `ASCENDER_REQUEST_PROFILE_WITH_DOT`, `ASCENDER_RUNNER_OMIT_ENV_FILES`, `ASCENDER_RUNNER_SUPPRESS_OUTPUT_FILE`.

Same route as #984: renamed at the definition and every use, with the old name added to the `_FORMER_NAMES` table, so a `conf.d` file that writes `AWX_NOTIFICATION_REQUEST_TIMEOUT` still lands on the current name.

## The four from the environment, which could not take that route

`AWX_SETTINGS_FILE`, `AWX_SETTINGS_DIR`, `AWX_LOGGING_MODE`, `AWX_WEB_PROCESS`.

The alias table runs **after** the settings modules have loaded, and two of these decide *which settings file loads at all*. An alias applied afterwards would be too late to matter. They read both names directly instead, through one helper, in the same shape as the `ASCENDER_` variables ascender-kit took in [its #72](https://github.com/ctrliq/ascender-kit/pull/72).

## Nothing has to move

- `AWX_SETTINGS_FILE` is what the Dockerfile passes to `collectstatic`.
- `AWX_LOGGING_MODE` is what the three installers write into their compose environments.
- `AWX_WEB_PROCESS` is what supervisor sets.

Those keep working for as long as they exist, not for a release.

## Checking

Both directions against a real production settings load:

- the old environment name still selects the settings file, and
- a settings file written with the old setting name still carries over (`AWX_NOTIFICATION_REQUEST_TIMEOUT = 99` arrives as `settings.ASCENDER_NOTIFICATION_REQUEST_TIMEOUT == 99`).

A test asserts that no settings module reads one of the four with a bare `os.environ.get`, since that would answer to one name only, which is the failure the helper exists to prevent.

1,573 unit and conf tests pass. `typed.py` regenerated. `ruff check` and `ruff format --check` clean.

Touches `defaults.py` and `production.py`, so it conflicts with #997; whichever lands second is a rebase onto the moved paths.


blaipr:
Rebuilt on the post-#997 main: mergeable now, full suite green at 4238 passed, 6 skipped.

cigamit (approved):

PR #1012, opened 2026-09-15
fix: saving a workflow fails on a node added and deleted before the first save

Aureliolo:
##### SUMMARY

In the workflow visualizer, adding a node, editing it, then deleting it and saving
fails the save. The node was never posted, so there is nothing to update, but the
branch that decides what to do with each node cannot tell that case apart from an
ordinary edit.

`Visualizer.tsx` sorts every node into one of three cases:

```ts
if (node.isDeleted && node.originalNodeObject) {          // exists server side, removed
} else if (!node.isDeleted && !node.originalNodeObject) { // never posted
} else if (node.isEdited) {                               // edited
```

`originalNodeObject` is the API object a node was built from. `generateNodes` sets it
for nodes read from the API (`workflowReducer.ts:758`); `createNode` (`:480`) does not,
and a new node only gets one from the response of the save that posts it. `deleteNode`
(`:698`) and `deleteAllNodes` (`:559`) set `isDeleted` on whatever node they are handed,
and `updateNode` (`:1070`) sets `isEdited` the same way. None of them knows or cares
whether the node exists server side.

A node added, edited and deleted between two saves is therefore deleted, unposted and
edited at once. It fails the first branch for want of an `originalNodeObject`, fails the
second because it is deleted, and falls into the third, which patches a node the API has
never seen.

The fix requires `originalNodeObject` on the update branch. A node that was never posted
and is now deleted needs no request, so it matches no branch, which is the right outcome.

##### ISSUE TYPE

##### COMPONENT NAME
 - UI

##### ASCENDER VERSION
```
25.6.3.dev176+ge18470b53
```

##### ADDITIONAL INFORMATION

**Steps to reproduce**

1. Open a workflow job template, Visualizer.
2. Add a node, any job template will do.
3. Open that node and save the node modal, so it counts as edited.
4. Delete that node.
5. Save.
6. Before: `Error saving the workflow!`. After: the save completes.

"Delete All Nodes" reaches the same branch if any of the nodes was added and then
edited in the current session.

**What the failure looks like**

On main the patched id is `undefined`:

```
PATCH /api/v2/workflow_job_template_nodes/undefined/ -> 404
```

`Base.update()` interpolates the id straight into the URL, so nothing catches it earlier.

Before the TypeScript migration the same branch read `node.originalNodeObject.id` and
the save handler threw. That is what it looks like on the released 25.6.0 UI, which is
where I ran into it:

```
TypeError: Cannot read properties of undefined (reading 'id')
    at main.<hash>.js:6:2292709
    at Array.forEach (<anonymous>)
    at main.<hash>.js:6:2290771
    at main.<hash>.js:2:222417
    at onSave (main.<hash>.js:6:2298678)
```

The migration wrote the access as `node.originalNodeObject?.id as number`, which turned
the throw into the bogus PATCH. The `?.` is what the type checker needs to accept an
optional field, so `tsc` has nothing to say here: the case is hidden rather than handled.

**Partial save**

Either shape fails after work has already been done. Node requests are dispatched as the
loop builds them, so on main every node change is applied and then the 404 rejects
`Promise.all(nodeRequests)`; before the migration the throw stopped the loop and applied
whatever it had reached. In both cases everything queued behind the node loop, link
association and disassociation, approval templates, credentials, labels and instance
groups, is skipped. The visualizer reports a save error over a workflow that has moved
half way, and a reload shows node changes without the link changes that went with them.

**Test evidence**

Before, the new test against the unfixed component:

```
FAIL  src/screens/Template/WorkflowJobTemplateVisualizer/Visualizer.test.tsx > Visualizer > Successfully saves a node added and deleted in the same session
AssertionError: expected "vi.fn()" to be called with arguments: [ 9000 ]

Number of calls: 0

 Test Files  1 failed (1)
      Tests  1 failed | 13 skipped (14)
```

The API client is automocked, so `update` returns `undefined` and the `.then` chained
onto it throws inside the loop. The node that does exist server side is never reached
and `destroy` is never called. Against the real client that same call is the 404 above.

After:

```
 ✓ Visualizer > Successfully deletes all nodes
 ✓ Visualizer > Successfully saves a node added and deleted in the same session
 ✓ Visualizer > Error shown when saving fails due to node edit error

 Test Files  1 passed (1)
      Tests  14 passed (14)
```

```
$ TZ='UTC' npx vitest run src/screens/Template src/components/Workflow
 Test Files  60 passed (60)
      Tests  349 passed (349)

$ npx tsc --noEmit
$ npx eslint src/screens/Template/WorkflowJobTemplateVisualizer/
$ npx prettier --check src/screens/Template/WorkflowJobTemplateVisualizer/
All matched files use Prettier code style!
```

**Design note**

The guard could go on the second branch instead, restructuring to `if (node.isDeleted)`
with the destroy nested inside, which would make "deleted and never posted" explicit
rather than a fall-through to nothing. The one-line form was kept because the three
branches are already written as a chain of independent predicates, and a reader checking
the update branch in isolation now sees the condition it depends on.

**Not addressed here**

The `?.` and `as number` casts on `node.originalNodeObject` inside the update branch are
redundant at the point of the guard, but the narrowing does not survive into the `.then`
callbacks, which reassign `node.originalNodeObject`. Left alone to keep the diff to the
behaviour change. Happy to fold that in if you would prefer it here.


cigamit (approved):
Thanks!

PR #1015, opened 2026-09-15
feat: warn when a workflow limit replaces the limit set on a node

Aureliolo:
##### SUMMARY

A workflow job template's limit replaces the limit of every node whose template prompts for one (`node_prompts_data.update(wj_prompts_data)`, `ascender/main/models/workflow.py:571`). The node modal still takes and saves a node limit without any sign that it will not be used.

This adds an inline warning under the limit field:

- in the node modal, when the workflow job template has a limit of its own
- in a workflow's launch prompt, the way `useInventoryStep` already warns for an inventory

The visualizer passes its template to the node modals through a new `WorkflowTemplateContext`.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - UI

##### ASCENDER VERSION
```
25.6.3.dev178+g33d8a6aa3
```

##### ADDITIONAL INFORMATION

```
$ TZ='UTC' npx vitest run src/screens/Template src/components/Workflow src/components/LaunchPrompt
 Test Files  70 passed (70)
      Tests  404 passed (404)
```

`tsc --noEmit`, `eslint` and `prettier --check` are clean.


cigamit:
Just needs a rebase / regeneration of the po files

Aureliolo:
Rebased onto main and regenerated the po files.

cigamit (approved):

PR #1031, opened 2026-09-17
refactor: name the task manager toggle ascender

blaipr:
`AWX_DISABLE_TASK_MANAGERS` was missed by the settings rename: it is absent from `_FORMER_NAMES` in `production.py`, and still defined in `development.py` and read in `api/views/debug.py` and `main/scheduler/tasks.py` under the old name, messages included.

It is development only, guarded by `MODE == 'development'`, and it is the kind of setting that lives in someone's own `local_settings.py` rather than in this repository. So the rename cannot reach where it is actually written, and the old name is read after the override chain.

Dropping it would not have failed loudly. The toggle would have been ignored and the task managers would have kept running, which for a debugging switch reads as the debugger being broken rather than as a renamed setting.

## The precedence rule, and its limit

The old name applies while the current one is still at its default. That is the same rule as `_FORMER_NAMES`, and it carries the same limitation, which the comment now states outright: a file setting `ASCENDER_DISABLE_TASK_MANAGERS = False` explicitly cannot be told apart from a file not mentioning it, because the default is also `False`. So writing both names with conflicting values resolves to the old one. Verified all four combinations:

| local settings | result |
| --- | --- |
| neither set | `False` (default) |
| old name `True` | `True` |
| new name `True` | `True` |
| old `True`, new `False` | `True` (the documented limit) |

## AWX_KUBE_DEVEL

Renamed the same way while it is in reach. Nothing in this tree sets it, it arrives from outside, so `development.py` and the three `launch_awx_*.sh` scripts accept either name rather than assuming the new one shows up.


blaipr:
Fixed in b03892e: truthiness before the fallback, so the settings and the launch scripts now agree on all nine combinations of unset, empty and set, including the empty-new-name-with-old-name-set case that was the disagreement.

cigamit (approved):

PR #1032, opened 2026-09-17
perf: warm the popular screens on idle instead of on the click

blaipr:
An alternative to #1017, for @cigamit to compare against. Either fixes the first click wait; this one keeps the entry bundle small while doing it.

The lazy split in #939 was mine, so the sluggishness is mine to answer for.

## The idea

`import()` populates the module registry. Warm the five most used screens during idle once the authenticated shell has painted, and when the route renders `React.lazy` resolves from the registry with no fetch and no `Suspense` fallback.

Same user visible result as a static import. The screens just never enter the bundle everyone downloads before the login form.

## Measured

All three built locally on node 24:

| | main | #1017 | this PR |
| --- | --- | --- | --- |
| entry chunk (gzip) | 108 KB | 316 KB | **108 KB** |
| JS chunks | 118 | 69 | **118** |
| entry cost vs main | — | **+208 KB** | **+197 bytes** |
| first click, those 5 | one fetch | none | **none** |
| first click, other 21 | one fetch | one fetch | one fetch |

The 197 bytes is the prefetch function. All 118 chunks stay split and the five screens remain their own chunks.

## It also helps Devel

The ~1s @cigamit reported is Vite transforming modules on demand, with no bundling. This moves that transform into idle time rather than the click, so it helps dev as well as production.

## Details worth knowing

- Called from `RenderAppContainer`, so nothing is warmed on the login page: there is no session to open a screen with until that mounts.
- `requestIdleCallback` with a `timeout: 3000`, because a busy tab may never reach idle, and that is exactly the machine where click latency is worst.
- Skipped on `saveData` or a `2g` `effectiveType`, so nobody's metered connection pays for a screen they may not open.
- A failed prefetch is swallowed: the route still loads it on demand.

## Tests

Five, covering idle scheduling, the timer fallback, both connection guards, and the fast connection case. Each was checked against a deliberately broken implementation:

- removing the `saveData` and `2g` guards fails the two connection tests
- removing the idle `timeout` fails the scheduling test

`tsc --noEmit` is clean across the project, `eslint` clean on the three touched files, and `routeConfig` plus `App` tests pass, 17 in total.

## The trade-off, stated plainly

Prefetch spends bandwidth on screens a session may never open, roughly 100 KB gzipped for the five, at idle and after paint. That is the one axis where #1017 is not worse, since it spends the same bytes anyway, just earlier and for everyone.

What is measured here is build cost. I have not profiled click latency in a browser, so the no-fetch claim rests on module registry semantics rather than a trace.


cigamit:
I'm just sitting on this until I get a chance to test in Production, so I can see if its even needed.  I don't think it is, as things are loaded differently there, and I don't believe the slowness will be as apparent.

PR #1037, opened 2026-09-20
feat: match LDAP users to teams by attribute, not just by group

fernandorocagonzalez:
##### SUMMARY

An LDAP org or team map can only name group DNs. The list in `users:` is compared against the groups of whoever is logging in, so granting one person access means asking whoever runs the directory for a group of one. That is a ticket to another team and a group nobody ever cleans up.

The social auth path has not had this problem for years. `SOCIAL_AUTH_*_TEAM_MAP` matches on username, email or a regular expression for SAML, OIDC, Azure AD, GitHub and Google. The same configuration being unwritable for LDAP looks like an inconsistency rather than a decision, and it is the first reason for this change.

Rather than invent a third syntax, this takes the one the platform already uses. `ansible_base.authentication` evaluates its authenticator maps against a trigger rule, and the half of that which answers "does this rule apply to this user?" is a handful of pure functions over plain dicts, with no models behind them. Those are vendored beside the rest of the django-ansible-base subset this repo already carries, and the existing settings based maps gain a key that feeds them:

```yaml
AUTH_LDAP_TEAM_MAP:
  Patching Viewers:
    organization: Example Org
    triggers:
      attributes:
        mail:
          equals: someone@example.com
    remove: true
```

Organization maps take one rule per role, `triggers_admins`, `triggers_users` and `triggers_auditors`, matching the `remove_<role>` keys already beside them. The operators are the platform's: `equals`, `matches`, `contains`, `ends_with` and `in`, over `attributes`, `groups`, `always` and `never`, joined with `and` or `or`.

Writing it this way means a rule someone writes today is the same JSON that `AuthenticatorMap.triggers` takes, so it carries over rather than being thrown away if and when this rebases onto the platform's authenticators.

##### Design notes

**Nothing existing moves.** `_update_m2m_from_map` delegates to `_update_m2m_from_groups` unchanged when there is no rule, which is every configuration that exists. A rule that says nothing about a user falls through to the group DNs beside it. There are tests pinning the old shape, including the bare string and boolean forms of `users:`.

**Attributes are the directory's own**, matched through the `ldap.cidict` django-auth-ldap decodes the entry into, so their names behave the way the protocol says they do. Populating the user has already read the entry, so matching on `department` or `title` costs no extra query.

**Group DNs are folded the way `is_member_of` folds them.** django-auth-ldap keeps DNs in their lower case normal form, so a DN written the way the directory prints it keeps working when it moves from `users:` into a rule.

**A rule that could only be half applied, or could not match anyone, is refused** rather than saved: more than one trigger type, more than one group operator, an empty rule body, or a pattern that does not compile. The evaluator honours one of each and ignores the rest, and with `remove` set a rule matching nobody strips the role from the whole directory, so these are worth catching at the API rather than in a log. The same check runs again at evaluation time, because these settings can be written to a settings file without ever passing through the serializer, and a rule that fails it decides nothing at all rather than revoking.

**The second commit is a separate bug.** The loop over the three organization roles assigns `remove_<role>` back to the same name the next role reads its default from, so `remove_admins: false` quietly stops members and auditors being removed too. `social_pipeline.py` reads the same options into a `role_remove` local and has always got this right. It is split out because it is a behaviour change on its own, and it is in this PR because `remove` is now what decides whether an unmatched rule revokes.

##### What is vendored

`ascender/dab/authentication/utils/`, following the procedure in `ascender/dab/VENDORED.md`: `trigger_definition.py` verbatim, the trigger evaluation half of `claims.py`, and `_validate_trigger_data` lifted out of `serializers/authenticator_map.py`, where it only used `self` to recurse. The models, migrations, serializers, views and the eighteen authenticator plugins are all left upstream. Two local modifications, both documented there and both covered by tests:

- `_is_case_insensitivity_enabled` reads the `AUTH_MAP_CASE_INSENSITIVE` setting, since django-flags is not a dependency here.
- `process_user_attributes` copies before popping `join_condition`. Upstream hands it a dict freshly loaded from a `JSONField` every time so it never notices, but ours comes from the settings cache and would lose the join after the first login.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API
 - Docs

##### ASCENDER VERSION
```
awx: 25.6.3.dev203+g6363acd5b
```

##### ADDITIONAL INFORMATION

`docs/auth/ldap.md` gains a section on how the maps are read and what a rule can say.

Tests: 70 in `ascender/sso/tests/functional/test_backends.py`, three of which drive `on_populate_user` end to end; 74 in `ascender/sso/tests/unit/test_fields.py`; and 15 in `ascender/main/tests/functional/test_dab_authentication_triggers.py` covering the vendored evaluator and the two local modifications.

No migrations. `AUTH_MAP_CASE_INSENSITIVE` is a plain Django setting, so it does not appear in the settings API; it is in `ascender/settings/typed.py` because `ascender/conf/tests` asks for it. `ascender/ui/src/types/api.generated.ts` is updated for the two help texts that changed, since CI regenerates it and diffs.


fernandorocagonzalez:
The red CI on the earlier run here is not this change. Rebased onto current main now, and worth writing down what the four failures were, since the same three are on every open pull request:

**`api-lint`, `ui-lint` and `ui-api-types-drift` come from #1036**, which removed Pendo without reformatting `ascender/conf/registry.py`, without reformatting the three files it edited under `ascender/ui/src/screens/Setting/UI/`, and without rerunning `make ui-api-types`. The evidence that it is not this branch: a dependabot pull request that touches only `.github` fails the same three jobs and `ui-lint` names the same three files, and `ascender/conf/registry.py` passes `ruff format --check` at dc68e323 and fails at de011214. #1049 fixes all three.

**`api-migrations`** never ran. Its annotation is "the hosted runner lost communication with the server" and the step has no conclusion.

**`api-test` passed**, which is the job that runs the tests this change adds.

While looking at the drift job I used its output to check the one thing about this branch I could not verify locally: `api.generated.ts` is edited by hand here, because regenerating it needs a dev environment. None of the lines this branch touches appear in the diff that job prints, so the committed copy matches what regeneration produces for the two help texts that changed.

Both Copilot findings are fixed in fbdf3318, now 5f15f277 after the rebase. The empty operand one was real and slightly wider than reported: `has_and: []` and `has_not: []` match every user, and so do `contains: ""`, `ends_with: ""` and `matches: ""`, while `has_or: []` and `in: []` match none, so with `remove` set either answer reaches the whole directory. All of them are refused now, along with a second operator on one attribute, which the evaluator would have dropped silently.

cigamit (approved):

